Issue #4, opened 2021-05-21
move "bad: missing PASETO token" to helpers.go


liaohongxing:
https://github.com/gofiber/contrib/blob/dab972f859dd3d9a2032b5c86780c667f833ee2a/paseto/paseto.go#L35

errors.New("bad: missing PASETO token")   moved to helpers.go#L22 ，so that "errors.Is"

https://github.com/gofiber/contrib/blob/dab972f859dd3d9a2032b5c86780c667f833ee2a/paseto/helpers.go#L22

@Jictyvoo

Jictyvoo:
I don't understand your issue well. You requesting to move this error to `helpers.go`?

liaohongxing:
example code
```
ErrorHandler: func(c *Ctx, err) error {
   if errors.Is(paseto.ErrMissingToken) {
      return c.JSON(fiber.Map{"MissingTokenLang": "I18n"});
      // or
      return c.Redirect("/signin")
   }
}
```

Jictyvoo:
I see, ok, I gonna make a pull request to improve this

PR #5, opened 2021-05-21
PASETO: Replace Payload struct by paseto.JSONToken + Improve helpers and public errors

Jictyvoo:
As requested in #4 this PR improve some things about config and move all errors to helpers.go

- All errors are now public
- Improve default ErrorHandler error verification
- Add ContextKey and TokenLookup configs to DefaultConfig

More additions
- Replaced `Payload` struct by `paseto.JSONToken`  struct
- Fixed some typos
- Use `ConfigDefault` values when a given config is empty, to be more consistent with `ConfigDefault`

### DIFF
--- a/paseto/config.go
+++ b/paseto/config.go
@@ -7,10 +7,11 @@ import (
 	"time"
 
 	"github.com/gofiber/fiber/v2"
+	"github.com/o1egl/paseto"
 	"golang.org/x/crypto/chacha20poly1305"
 )
 
-// Config defines the config for BasicAuth middleware
+// Config defines the config for PASETO middleware
 type Config struct {
 	// Filter defines a function to skip middleware.
 	// Optional. Default: nil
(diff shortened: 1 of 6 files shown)


Jictyvoo:
I've update in another branch to replace the Payload struct and utilize the `JSONToken` in paseto module. I add this update in this PR or create another one?

ReneWerner87:
here is the right place for the tests that are important for your customization, since the pull request is not closed yet, you should add them here

Jictyvoo:
@ReneWerner87 I think it's done now.
Can you review the change of our custom Payload object to use the paseto.JSONToken?

Also, I'm thinking on use `Storage` to stores the uuid and check this ID in every token validation on this middleware, and to disable it, just don't pass any storage to the `Config`, what you think?

ReneWerner87:
@Jictyvoo hmm, you said you are through, however i don't see any new testcases for the new errors ?

Jictyvoo:
> @Jictyvoo hmm, you said you are through, however i don't see any new testcases for the errors ?
> 
> can you please include a test for these different errors where you check the correct misbehavior
> 
> some should already exist, some not yet
> 
> ```go
> ErrExpiredToken  = errors.New("token has expired")
> ErrMissingToken  = errors.New("missing PASETO token")
> ErrDataUnmarshal = errors.New("can't unmarshal token data to Payload type")
> fiber.StatusBadRequest
> ```

Added some tests to cover that. But to do that I detach the defaultErrorHandler to be a const function instead of a anonymous function

Also, what you think about the change of `Payload` struct to `paseto.JSONToken`?

ReneWerner87:
sorry forgot to answer your question

> Also, what you think about the change of Payload struct to paseto.JSONToken?

is currently okay for me, if we get problems with it, we can still change it, software grows as you use it

Jictyvoo:
> sorry forgot to answer your question
> 
> > Also, what you think about the change of Payload struct to paseto.JSONToken?
> 
> is currently okay for me, if we get problems with it, we can still change it, software grows as you use it

No problem. Now I gonna think if it's good to add a Storage parameter to the middleware to stores token/uuid to add a verification to it

PR #56, opened 2022-01-13
update otel

### LINKED ISSUE #47: OTel: add method to span name
Currently, otel middleware [sets](https://github.com/gofiber/contrib/blob/97ea1c071d84861407b9867b0d569c6a70ee27c2/otelfiber/fiber.go#L84) span name to route path. However, it's possible to register multiple routes with the same path but different methods. For example, `POST /orders` and `GET /orders`.
I believe, span name should include method along with path.

### LINKED ISSUE #48: OTel: tracer name looks strange
It's [recommended](https://opentelemetry.io/docs/instrumentation/go/getting-started/#trace-instrumentation) to set tracer name to the package name. However, currently, tracer name is `go.opentelemetry.io/contrib/instrumentation/github.com/gofiber/fiber/otelfiber`.

Perhaps, it should be `github.com/gofiber/contrib/otelfiber`

### LINKED ISSUE #58: Otel: the instrumentation sets empty `http.server_name`
According to the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions):

> This should be obtained via configuration. If no such configuration can be obtained, this attribute MUST NOT be set

However, this instrumentation always sets `http.server_name` even when the provided server name is empty

WLun001:
- fix #47 
- fix #48 
- fix #58
- update to include `EnduserIDKey`
- update example


otelfiber/fiber.go
```diff
@@ -16,7 +19,7 @@ import (
 
 const (
 	tracerKey  = "otel-go-contrib-tracer-gofiber-fiber"
```
efectn, line 21: I think you should update tracerKey also.

WLun001: @efectn what is your suggestion? I am borrowing from https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/labstack/echo/otelecho/echo.go#L32

efectn: fiber-contrib-tracer-gofiber-fiber 🤔

efectn (commented):
May you add README like https://github.com/gofiber/contrib/blob/main/fiberzap/README.md, https://github.com/gofiber/contrib/tree/main/fibersentry?

WLun001:
@efectn not sure what to add to the readme, as we already have a minimum example code

efectn (approved):
Looks good to me.

WLun001:
@efectn updated with 2 more fixes

efectn:
Do you think to include metrics into the otelfiber? I think it would be great i think.

WLun001:
@efectn metrics leave it for next time

WLun001:
Good to merge? @efectn

PR #59, opened 2022-01-19
add requestId

wanghao0123:
add request id to track every request


efectn (changes requested):
Can you add requestId into the unit tests?

wanghao0123:
> Can you add requestId into the unit tests?

sure, here you are

efectn (approved):

PR #60, opened 2022-01-20
fix spell refeer -> referer

piyongcai:
only fix refeer to referer


ReneWerner87:
@efectn @piyongcai 
Unexpected that there is a bug, we should add more unit tests for these cases, then this bug would not even be in the last release version

PR #65, opened 2022-01-29
Swagger middleware added

ziyaozclk:
The added middleware handles the Swagger JSON file. I use [Fiber](https://github.com/gofiber/fiber) in many projects and I need it very much. I'm sure it will be useful to many people.


efectn (changes requested):
Can you add Swagger to README.md

efectn (approved):

ReneWerner87 (approved):

PR #191, opened 2022-05-26
fix(190): check if is a Basic header

### LINKED ISSUE #190: Otelfiber should change how to deal with the Authorization header.
When otelfiber is checking the `Authorization` header it will always panic if the provided header doesn't have at least 5 of length.

Normally this is acceptable, but to prevent a unnecessary panic, we should check the length first and return `"", false` if the length is invalid.

`fiber.go:128`
```go
	// Decode the header contents
	raw, err := base64.StdEncoding.DecodeString(auth[6:])
	if err != nil {
		return "", false
	}
```

lucasoares:
Fixes #190

### DIFF
--- a/otelfiber/fiber.go
+++ b/otelfiber/fiber.go
@@ -3,9 +3,10 @@ package otelfiber
 import (
 	"context"
 	"encoding/base64"
-	"github.com/gofiber/fiber/v2/utils"
 	"net/http"
 	"strings"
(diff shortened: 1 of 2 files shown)


ReneWerner87 (approved):

lucasoares (commented):
LGTM? @ReneWerner87

PR #263, opened 2022-08-15
Feature/opafiber

mstrYoda:
Add Open Policy Agent support as fiber middleware.


opafiber/fiber.go
```diff
@@ -0,0 +1,79 @@
...
+	return func(c *fiber.Ctx) error {
+		input := map[string]interface{}{
+			"method": c.Method(),
+			"path":   c.Path(),
+		}
+
+		if cfg.IncludeQueryString {
+			queryStringData := make(map[string][]string)
+			c.Request().URI().QueryArgs().VisitAll(func(key, value []byte) {
+				queryStringData[string(key)] = append(queryStringData[string(key)], string(value))
```
mstrYoda, line 53: Is that be more appropriate to use `utils.UnsafeString(value)` instead of `string(value)` here?

efectn: Sounds better. This won't be used outsuide of the handler. So we can use it i think. Fiber also uses unsafe when parsing the query.

mstrYoda: Thx, done 👍

mstrYoda:
@efectn What about this pr? Are we able to merge it?

efectn (approved):

.github/workflows/test.yml
```diff
@@ -52,3 +52,38 @@ jobs:
                 run: cd ./casbin && go test ./... -v -race
             -   name: Test fibernewrelic Middleware
                 run: cd ./fibernewrelic && go test ./... -v -race
+    Tests2:
```
ReneWerner87, line 55: instead of copying the whole thing, can't you just add more lines with an additional IF condition where you ignore the other run for 1.15?

https://github.com/gofiber/storage/blob/main/.github/workflows/test.yml#L52

https://docs.github.com/en/actions/learn-github-actions/expressions#startswith

`if: startsWith(matrix.go-version, '1.15') === false`

ReneWerner87:
after the adjustment with the workflow i will merge it

.github/workflows/test.yml
mstrYoda: Thank you for feedback, I updated the tests

ReneWerner87 (approved):

Issue #143, opened 2022-04-05
otelfiber: issue with Method


gzuidhof:
This line:

```go
semconv.HTTPMethodKey.String(c.Method()),
```

is problematic in real use: the method may be sent asynchronously to a collector/monitoring service outside of the request scope. It results in this arriving on the other end:
<img width="116" alt="Screenshot 2022-04-05 at 20 22 42" src="https://user-images.githubusercontent.com/1039510/161823898-bac7f4c7-7892-44f0-b55d-b688b5e1f73c.png">

We should be making a copy of the method.

PR #270, opened 2022-08-20
fix: use `UserContext()` & fix incorrect strings

GalvinGao:
This is to infer context from `UserContext()` instead, so the value set in `UserContext()` by users will not be simply overwritten and simply been discarded.

This PR also resolves issue #143.

---

Discussed in Discord

> Hi! I was trying to use https://github.com/gofiber/contrib/tree/v0.4.1/otelfiber and I noticed that my values set in UserContext() is not passing down. Further digging showed that at https://github.com/gofiber/contrib/blob/v0.4.1/otelfiber/fiber.go#L53-L56 the middleware just used Context() (which is of type *fasthttp.RequestCtx()) to create a new context of type context.Context and later overrides the UserContext() with that. 
>
> I think it would be more feasible for the middleware to infer context from UserContext() instead, so the value set in UserContext() will not be simply discarded.

### DIFF
--- a/otelfiber/fiber.go
+++ b/otelfiber/fiber.go
@@ -62,28 +62,31 @@ func Middleware(service string, opts ...Option) fiber.Handler {
 		ctx := cfg.Propagators.Extract(savedCtx, propagation.HeaderCarrier(reqHeader))
 		opts := []oteltrace.SpanStartOption{
 			oteltrace.WithAttributes(
+				// utils.CopyString: we need to copy the string as fasthttp strings are by default
+				// mutable so it will be unsafe to use in this middleware as it might be used after
+				// the handler returns.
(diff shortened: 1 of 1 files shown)


efectn (approved):

gaby (approved):
Looks good!

gaby:
@ReneWerner87 Can you take a look at this one. Thanks!

gzuidhof:
Thank you for this! Could you perhaps mint a new release that includes this change @ReneWerner87?

ReneWerner87:
Sure, will do it in some hour.

gaby:
Thanks! 😁

ReneWerner87:
https://github.com/gofiber/contrib/releases/tag/v0.5.0

PR #272, opened 2022-08-25
Add ability to ignore specific URI list, eg: health check /health

hotsnow:


hotsnow:
`
func HasPrefix(str,  substring) bool
`

Or maybe this to ignore the URI by prefix?

ReneWerner87 (approved):

PR #359, opened 2022-11-05
support configuring log levels and reqHeaders support

DoubleDi:
This PR has 2 features:
1. Supports dumping all the headers with the reqHeaders field
2. Allows to specify your own log levels in a similar way to log messages

As a bonus it speeds up the middleware as the fields are not getting marshalled if the log won't be sent to the output

### DIFF
--- a/fiberzap/config.go
+++ b/fiberzap/config.go
@@ -3,6 +3,7 @@ package fiberzap
 import (
 	"github.com/gofiber/fiber/v2"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 // Config defines the config for middleware.
@@ -41,6 +42,11 @@ type Config struct {
 	//
 	// Optional. Default: {"Server error", "Client error", "Success"}
 	Messages []string
+
+	// Custom response levels.
+	//
+	// Optional. Default: {zapcore.ErrorLevel, zapcore.WarnLevel, zapcore.InfoLevel}
+	Levels []zapcore.Level
 }
 
 // Use zap.NewProduction() as default logging instance.
@@ -52,6 +58,7 @@ var ConfigDefault = Config{
 	Logger:   logger,
 	Fields:   []string{"latency", "status", "method", "url"},
 	Messages: []string{"Server error", "Client error", "Success"},
+	Levels:   []zapcore.Level{zapcore.ErrorLevel, zapcore.WarnLevel, zapcore.InfoLevel},
 }
 
 // Helper function to set default values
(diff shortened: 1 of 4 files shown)


fiberzap/zap.go
```diff
@@ -81,8 +82,25 @@ func New(config ...Config) fiber.Handler {
...
+		switch {
+		case s >= 500:
+			ce = cfg.Logger.Check(cfg.Levels[0], cfg.Messages[0])
+		case s >= 400:
+			ce = cfg.Logger.Check(cfg.Levels[1], cfg.Messages[1])
+		default:
+			ce = cfg.Logger.Check(cfg.Levels[2], cfg.Messages[2])
+		}
+
+		if ce == nil {
```
efectn, line 97: Do we really need this?

efectn:
Can you check my comment @DoubleDi 🚀

fiberzap/zap.go
DoubleDi: the Check func will return a nil logger if the level is lower, than the configured logger level, so yes, we need a check here.

DoubleDi:
@efectn Hi! Responded!

fiberzap/zap.go
efectn: Oh right

efectn (approved):

fiberzap/config.go
```diff
@@ -41,6 +42,11 @@ type Config struct {
 	//
 	// Optional. Default: {"Server error", "Client error", "Success"}
 	Messages []string
+
+	// Custom response levels.
+	//
+	// Optional. Default: {zapcore.ErrorLevel, zapcore.WarnLevel, zapcore.InfoLevel}
+	Levels []zapcore.Level
```
ReneWerner87, line 49: please add something in the readme in the config section for this

fiberzap/zap.go
```diff
@@ -81,8 +82,25 @@ func New(config ...Config) fiber.Handler {
 			stop = time.Now()
 		}
 
+		// Check if the logger has the appropriate level
+		s := c.Response().StatusCode()
+		var ce *zapcore.CheckedEntry
+		switch {
+		case s >= 500:
+			ce = cfg.Logger.Check(cfg.Levels[0], cfg.Messages[0])
```
ReneWerner87, line 90: what happens if i change the series followers of the registered log level e.g.
```go
Levels: []zapcore.Level{zapcore.InfoLevel, zapcore.WarnLevel, zapcore.ErrorLevel},
```
or i remove some
```go
Levels: []zapcore.Level{zapcore.ErrorLevel},
```


currently, the construct does not seem safe to me

- [ ] please also add some tests for it

ReneWerner87 (changes requested):
need some changes

fiberzap/config.go
DoubleDi: done

ReneWerner87:
look at it at the weekend

ReneWerner87 (approved):

PR #404, opened 2022-12-10
Deprecate support for Go 1.14, and 1.16

gaby:
* Update all go.mod files to use go `1.18`
* Update test workflows to use `1.18.x` and `1.19.x`

Script used to update everything: 

```bash
git clone git@github.com:gofiber/contrib.git
cd contrib/
// update every go.mod file to 1.18
for i in `ls -d */`; do cd $i; go mod tidy; go get -u; go mod tidy ; cd ../ ; done
```

Fixes #403 #399 #391 #392 #390 #387 #363


efectn (approved):

ReneWerner87:
Why should we do that? Does fasthttp specify a depreciation again?

ReneWerner87 (approved):

PR #385, opened 2022-11-29
feat(otelfiber): add metrics to otelfiber

ijsong:
This patch adds metrics to otelfiber and follows conventions defined by [Semantic Conventions for HTTP Metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#semantic-conventions-for-http-metrics). The metric instruments are added:

- http.server.duration
- http.server.request.size
- http.server.response.size
- http.server.active_requests

The Metrics package in the OpenTelmetry-Go is still in progress. However, I believe that this PR is a good starting point for otelfiber's metrics.

Resolves #50

PR #409, opened 2022-12-20
Update `otelfiber` metrics implementation

onematchfox:
Whilst trying out the metrics implementation added in #385 I noticed that all metrics were being reported on the route `/`. This is because the route was being accessed prior to `c.Next()` being called. As per the [docs](https://docs.gofiber.io/api/ctx#route) (and the existing tracing implementation) this can be only be set later in the method. 

> Do not rely on c.Route() in middlewares before calling c.Next() - c.Route() returns the last executed route.

So, first and foremost, this PR sets out to resolve that issue. In doing so I decided to use the opportunity to add the status code [attribute as per the recommendations](https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/http-metrics/#attributes) and, to prevent issues in future, expanded the tests to assert that the metrics are being produced as expected. 

On a related note, I noticed that:
- this module currently makes use of `v1.4.0` of the `semconv` package ([here](https://github.com/gofiber/contrib/blob/b9cfd37ac3c5dbc046e775875a625daf05bbfce3/otelfiber/semconv.go#L7) and [here](https://github.com/gofiber/contrib/blob/b9cfd37ac3c5dbc046e775875a625daf05bbfce3/otelfiber/fiber.go#L19)). Given that it is referencing the latest versions of the main libraries I wonder if this shouldn't also be updated to the latest version (`v1.12.0`) as well?
- related to the above, the attributes added this module (given it's inspiration on the `semconv` module) don't actually conform with the semantic conventions spec for either [metrics](https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/http-metrics/) and or [traces](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/). E.g. the use of `http.server_name` over `net.host.name`

FWIW, I did initially consider bundling changes to the attribute names to comply with the latest spec into this PR but figured those changes probably deserved a separate PR as I'm sure there is some discussion to be had.


gaby:
@onematchfox Seems like tests are failing

onematchfox:
> @onematchfox Seems like tests are failing

Ooops... My bad! Should all be working now.

gaby:
@onematchfox Can you bump the semconv version too? I ran into this issue about metrics being reported on / a couple days a go. So this PR should help with that.

gaby (changes requested):
Looks good, the only thing would be bumping the semconv version to .12

onematchfox:
> Looks good, the only thing would be bumping the semconv version to .12

Bumped version as requested. I have however left the metric attributes as is for now. Do you also want to tackle aligning these with the latest version of the spec as well in this PR?

For reference see
HTTP Metrics - https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/http-metrics/
HTTP Spans - https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/

gaby:
@onematchfox Given that we are using the latest opentelemetry, makes sense to also update the metrics attributes in this PR.

onematchfox:
> @onematchfox Given that we are using the latest opentelemetry, makes sense to also update the metrics attributes in this PR.

All done now. Suspect there's a few thing we may need to discuss :). Most pertinently, I wasn't quite sure how you wanted to handle the "removal" of `http.server_name`. Have tried to retain the attribute for those that depend on it, however I'm not sure how everyone will feel about the change in the function signature?

otelfiber/fiber.go
```diff
@@ -32,28 +30,28 @@ const (
 )
 
 // Middleware returns fiber handler which will trace incoming requests.
-func Middleware(service string, opts ...Option) fiber.Handler {
-	if service == "" {
-		service = defaultServiceName
-	}
+func Middleware(opts ...Option) fiber.Handler {
```
ReneWerner87, line 33: this is a breaking change or ?

onematchfox: > this is a breaking change or ?

Yes, this is definitely a breaking change. It's being made as `http.server_name` is no longer listed as an attribute in the semantic conventions for either HTTP metrics and/or spans. As mentioned,

> I have tried to retain the attribute for those that depend on it, however I'm not sure how everyone will feel about the change in the function signature?

FYI: If any consumers do depend on the attribute, it can be retained going forward by making use of the `WithServerName` option as shown in the tests at https://github.com/gofiber/contrib/blob/f3bf30b3165d1aaf29e501e54c6ca640687bbc04/otelfiber/fiber_test.go#L73-L78

Happy to discuss alternatives/options if you want to avoid this.

ReneWerner87 (approved):

ReneWerner87:
@gaby can you check again, after your approvement i will merge it

gaby:
Sure, will do!

gaby (approved):
Looks good, we may need to bump the release version since this is a breaking change when using `WithServerName()`

PR #425, opened 2023-01-11
Add support for passing existing New Relic application

### LINKED ISSUE #269: There should be an option to pass the existing NewRelic app without creating a new one
Greetings,

Currently, NewRelic middleware requires the configuration parameters to be passed and creates a new NewRelic application. However, on the project I am working on, I already have a NewRelic app and I use it some other places as well. I totally understand the basic-friendly approach, however, can we also have a function something like `fibernewrelic.NewWithApp(app *newrelic.Application)` that enables using the existing app? If so, I can start a PR that implements such a change.

gaby:
The main purpose of this PR is to fix #269 allowing the user to pass an existing `newrelic.Application`. If passed all the other config values will be ignored except for the transport type, since it's used by middleware function.

Also included:
- `TransportType` is now case insensitive. We now run ToUpper internally.
- Updated readme with example usage
- Added unit-test for testing different transport types (HTTP, HTTPS, http, invalid (defaults to HTTP))
- Added unit-test for testing against an invalid key (shorter than 40 chars)
- Added unit-test for passing a valid existing newrelic.Application
- Added unit-test for passing an invalid newrelic.Application

### DIFF
--- a/fibernewrelic/README.md
+++ b/fibernewrelic/README.md
@@ -22,13 +22,13 @@ fibernewrelic.New(config fibernewrelic.Config) fiber.Handler
 
 ### Config
 
-| Property       | Type         | Description                      | Default     |
-|:---------------|:-------------|:---------------------------------|:------------|
-| License        | `string`     | Required - New Relic License Key | `""`        |
-| AppName        | `string`     | New Relic Application Name       | `fiber-api` |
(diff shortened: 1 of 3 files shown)


efectn:
@mstrYoda can you check this ^^^

mstrYoda (approved):
/lgtm

ReneWerner87 (approved):

PR #441, opened 2023-02-02
change newrelic fiber to group transaction metrics under the same route

mstrYoda:
It prevents creating separate transactions on NewRelic.

It groups different path params under the same route now.

Here is the NewRelic screenshot:

image

### DIFF
--- a/fibernewrelic/README.md
+++ b/fibernewrelic/README.md
@@ -22,12 +22,13 @@ fibernewrelic.New(config fibernewrelic.Config) fiber.Handler
 
 ### Config
 
-| Property       | Type          | Description                      | Default     |
-|:---------------|:--------------|:---------------------------------|:------------|
-| License        | `string`      | Required - New Relic License Key | `""`        |
-| AppName        | `string`      | New Relic Application Name       | `fiber-api` |
-| Enabled        | `bool`        | Enable/Disable New Relic         | `false`     |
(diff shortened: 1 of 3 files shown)


ReneWerner87 (changes requested):
could you also check in the tests which values you set in txn ? otherwise nothing would break if there was an error/mistake there

think we should deposit there our expectation attitude also with, which data we expect within

ReneWerner87:
have 3 ideas after looking at the code of the newrelic package more closely

First, we could remove the "TransportType" config property, don't see where it is used

okay now to the possibilities:
1. we put a logger in the app and offer this possibility via the config and look, after the middleware process, what we have in the logger and expected to be transferred
https://github.com/newrelic/go-agent/blob/master/config.go#L39

2. we allow to set the app from outside via config (maybe useful if you have already initialized the newrelic app and want to use it for different routes) and mock the methods of the app we put in, so we can check the information we expect to get
https://github.com/gofiber/contrib/blob/main/fibernewrelic/fiber.go#L36-L37
(is already possible^^, then we could also use that and methods which are used internally for logging, overlay and collect data)

3. we make it possible to override the transport instance used in the newrelic app and give in this transport instance a listener for the http connections that should go out 
https://github.com/newrelic/go-agent/blob/master/config.go#L161
then with the info we have collected we can also check again the expectation of the flow

ReneWerner87 (approved):

PR #499, opened 2023-02-28
Migrate to OTEL-Go v1.14.0

kilianstallz:
OpenTelemetry-Go did change the sigature for the metric.Reader interface.
They also deprecated the Unit interface in favor of a string.

To keep using otelfiber with the latest otel-go packages we need to replace the Unit struct with string constants.
I also adapted the call signature of the reader interface like it is shown in the libraries tests.

```	
metrics := metricdata.ResourceMetrics{}
err := reader.Collect(context.Background(), &metrics)
```

### DIFF
--- a/otelfiber/fiber.go
+++ b/otelfiber/fiber.go
@@ -13,7 +13,6 @@ import (
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/metric/global"
 	"go.opentelemetry.io/otel/metric/instrument"
-	"go.opentelemetry.io/otel/metric/unit"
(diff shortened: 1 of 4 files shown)


ReneWerner87:
does this mean that it is a breaking change and we should upgrade the major version of the package ?

ReneWerner87:
> The usage of Otelfiber would not change, as the Middleware API did not change.

this is enough to know 👍

eko (approved):

eko:
Hi, do we expect anything else on this PR?

ReneWerner87 (approved):

PR #529, opened 2023-04-07
:rocket: Feature: Zerolog middleware

jcyamacho:
# Zerolog middleware

Add [zerolog](https://zerolog.io/) support to fiber, based on [fiberzap](https://github.com/gofiber/contrib/tree/main/fiberzap)

```go
package main

import (
    "github.com/gofiber/fiber/v2"
    "github.com/gofiber/contrib/fiberzerolog"
    "github.com/rs/zerolog"
)

func main() {
    app := fiber.New()
    logger := zerolog.New(os.Stderr).With().Timestamp().Logger()

    app.Use(fiberzerolog.New(fiberzerolog.Config{
        Logger: &logger,
    }))

    app.Get("/", func (c *fiber.Ctx) error {
        return c.SendString("Hello, World!")
    })

    if err := app.Listen(":3000"); err != nil {
        logger.Fatal().Err(err).Msg("Fiber app error")
    }
}
```


.github/workflows/release-drafter-fiberzerolog.yml
```diff
@@ -0,0 +1,19 @@
+name: Release Drafter Fiberzerolog
```
gaby, line 1: The tests.yml has to be modified to include the tests from this middleware.

Dependabot.yml also has to be updated

gaby (changes requested):
Can you add the missing files, so the CI can run?

gaby:
@ReneWerner87 @efectn My understanding is that Fiber v3 already uses ZeroLog by default. So this middleware would probably go away once v3 is released?

.github/workflows/release-drafter-fiberzerolog.yml
jcyamacho: done! also included in `security.yml` :)

.github/dependabot.yml
```diff
@@ -62,3 +62,9 @@ updates:
           - "🤖 Dependencies"
       schedule:
           interval: "daily"
+    -   package-ecosystem: "gomod"
```
gaby, line 65: Spacing may be wrong here

jcyamacho: can you please check if the problem is with `fiberi18n`,  I'm using the spacing like others, and not sure which should be changed?

gaby: @jcyamacho You are right, problem is the previous item!

gaby (approved):
lgtm!

.github/dependabot.yml
jcyamacho: if you don't mind I can fix it in this pr, just let me know

gaby: @jcyamacho Sure, thanks!!

jcyamacho: @gaby done! I'm also added a new config to allow to get a new logger for request, I think it will be useful for integrating opentelemetry tracing id to the logger, or any other custom value.

efectn:
> @ReneWerner87 @efectn My understanding is that Fiber v3 already uses ZeroLog by default. So this middleware would probably go away once v3 is released?

Doesn't use but possible to use it with default logger middleware of fiber

fiberzerolog/README.md
```diff
@@ -0,0 +1,66 @@
+# Fiberzap
```
efectn, line 1: Don't forget here

.github/workflows/release-drafter-fiberzerolog.yml
```diff
@@ -0,0 +1,19 @@
...
+on:
+    push:
+        # branches to consider in the event; optional, defaults to all
+        branches:
+            - master
+            - main
+        paths:
+            - 'fiberzerolog/**'
+jobs:
+    draft_release_fiberzap:
```
efectn, line 11: And here

fiberzerolog/README.md
jcyamacho: good catch, thanks!

.github/workflows/release-drafter-fiberzerolog.yml
jcyamacho: done! thanks!

efectn (approved):

ReneWerner87 (approved):

Issue #532, opened 2023-04-11
Error is not returned from otelfiber.Middleware() handler


calloway-jacob:
The middleware returned from `otelfiber.Middleware()` always returns `nil`, even if the next middleware in the stack returned an error.

When using multiple middlewares together, this can prevent a middleware earlier in the stack from also having access to the error. This is undesirable, for example, if an earlier-added logging middleware should also log the error (and which is outside of the fiber application's configured ErrorHandler).

ReneWerner87:
@eko can you help here

calloway-jacob:
I've just noticed that my PR introduces a regression in the test (line 146 if `fiber_test.go`

calloway-jacob:
I suppose it is by design that this middleware catches any errors from middleware up the stack and directly invoke's the app's error handler? I see from the code comments that this is required to get the HTTP response status code to add into the tracing.

If that is the intended design, I don't see a way to make this compatible with other middlewares lower in the stack receiving access to the error additionally.

In that case, perhaps can we document that the tracing middleware should be the first one added to the stack?

calloway-jacob:
After working quite some time trying to combine the use of `fiberzap` and `otelfiber`, I've come to the conclusion that these two package middlewares are simply not 100% compatible with one another. As both of them call the app's error handler, it is necessary to return a nil error in each package.

I have an idea of how they could be made compatible, but it would require changes in each package as well as in fiber itself, which seems unlikely to be an accepted change.

The idea itself would be to introduce a new type, `HandledError`. This would represent an error that has already been passed to the app's error handler.

Any middleware that needs to call the app's error handler (and usually return a nil error), would have to check if the error is a `HandledError`. In that case, the middleware knows that the app's error handler was already called, and can get the response code and use the error for logging/tracing/etc.. In the case that an error is not a `HandledError`, the middleware is free to call the app's error handler, but must then return a `HandledError` (wrapping the original error) to signal to other middlewares that the error handler was already called for it.

Fiber itself would also have to take part in this scheme, and avoid calling the app's error handler when a middleware returns a `HandledError`.

I doubt that the payoff would be worth it here, but it would definitely extend the ability of multiple middlewares to react to an error up the stack without the fear of accidentally running side-effects from the error handler multiple times.

PR #533, opened 2023-04-11
Fix Issue 532

calloway-jacob:
This patch would fix issue #532, which I just opened.

### DIFF
--- a/otelfiber/fiber.go
+++ b/otelfiber/fiber.go
@@ -115,7 +115,8 @@ func Middleware(opts ...Option) fiber.Handler {
 		c.SetUserContext(ctx)
 
 		// serve the request to the next middleware
-		if err := c.Next(); err != nil {
+		err := c.Next()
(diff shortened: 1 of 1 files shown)


ReneWerner87 (approved):

gaby (approved):

gaby:
@calloway-jacob Can you take a look at the error in the tests?

calloway-jacob:
Hi @gaby, I wrote a bit about why I expect that this change cannot work with the existing code: https://github.com/gofiber/contrib/issues/532#issuecomment-1506488057

In short, middleware that calls the application error handler must return a nil error (as this middleware does), and it would be non-trivial to change this. That is why my proposed change breaks `TestErrorOnlyHandledOnce`.

Now that I understand it more fully, I'm happy to close this PR.

gaby:
@calloway-jacob Thanks for the info, Should we close the issue too?

PR #563, opened 2023-04-29
:rocket: Add minify contrib

alwismt:
# 🚀 Add minifying  HTML, CSS and JavaScript responses

## Description

This PR adds a new Minify middleware for [Fiber](https://github.com/gofiber/fiber). The middleware handles minifying HTML, CSS and JavaScript responses. This is located in the `minify` directory and contains the following files:

- `config.go`: defines the configuration options for the middleware.
- `css_minify.go`: contains the logic for minifying CSS responses.
- `html_minify.go`: contains the logic for minifying HTML responses.
- `js_minify.go`: contains the logic for minifying JavaScript responses.
- `minify.go`: contains the main middleware logic.
- `css_minify_test.go`, `html_minify_test.go`, `js_minify_test.go`, `minify_test.go`: contain test cases for each file and every possible test case.

### Related 
[🚀 Feature: Add minify middleware](https://github.com/gofiber/fiber/pull/2442)

## Changes:
- Added new middleware for minifying HTML, CSS and JavaScript responses.
- Created a new `minify` directory in the `root` directory and added the necessary files.
- Added configuration options for enabling/disabling minification for HTML, CSS and JavaScript responses.
- Added configuration options for excluding specific URLs and styles/scripts from minification.
- Added default configuration options and a helper function for setting default values.
- Added test cases for each file and possible test cases.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Usage:

```go
app.Use(minify.New())
```

### Custom Config

```go
cfg := minify.Config{
    MinifyHTML:        true,
    MinifyHTMLOptions: MinifyHTMLOptions{
      MinifyScripts: true,
      MinifyStyles:  true
    },
    MinifyCSS:         true,
    MinifyJS:          true,
    Method:            GET,
}

app.Use(minify.New(cfg))
```
### Config
| Property | Type | Description | Optional | Default |
| :--- | :--- | :--- | :--- | :--- |
| MinifyHTML | `bool` | Enable / Disable Html minfy | `yes` | `true` |
| MinifyHTMLOptions | `MinifyHTMLOptions` | [Options for the MinifyHTML](#MinifyHTMLOptions) | `yes` | `MinifyHTMLOptions` |
| MinifyCSS | `bool` | Enable / Disable CSS minfy | `yes` | `false` |
| MinifyCSSOptions | `MinifyCSSOptions` | [Options for the MinifyCSS](#MinifyCSSOptions) | `yes` | `MinifyCSSOptions` |
| MinifyJS | `bool` | Enable / Disable JS minfy | `yes` | `false` |
| MinifyJSOptions | `MinifyJSOptions` | [Options for the MinifyJS](#MinifyJSOptions) | `yes` | `MinifyJSOptions` |
| Method | `Method` | Representation of minify route method | `yes` | `GET` |
| Next | `func(c *fiber.Ctx) bool` | Skip this middleware when returned true | `yes` | `nil` |

### MinifyHTMLOptions
| Property | Type | Description | Default |
| :--- | :--- | :--- | :--- |
| MinifyScripts | `bool` | Whether scripts inside the HTML should be minified or not | `false` |
| MinifyStyles | `bool` | Whether styles inside the HTML should be minified or not | `false` |
| ExcludeURLs | `[]string` | URLs Exclud from minification | `nil` |

### MinifyCSSOptions
| Property | Type | Description | Default |
| :--- | :--- | :--- | :--- |
| ExcludeStyles | `[]string` | Styles exclud from minification | `[]string{"*.min.css", "*.bundle.css"}` |

### MinifyJSOptions
| Property | Type | Description | Default |
| :--- | :--- | :--- | :--- |
| ExcludeScripts | `[]string` | Styles exclud from minification | `[]string{"*.min.js", "*.bundle.js"}` |


minify/go.mod
```diff
@@ -0,0 +1,27 @@
+module github.com/gofiber/minify
+
+go 1.20
```
gaby, line 3: This should be go 1.18 or 1.19, Fiber supports 1.17

minify/minify_test.go
```diff
@@ -0,0 +1,594 @@
+package minify
```
gaby, line 1: The github actions CI file for running these tests is missing.

minify/minify_test.go
```diff
@@ -0,0 +1,594 @@
...
+
```
gaby, line 2: Add an entry for this middleware in .github/dependabot.yml

gaby (changes requested):

minify/minify_test.go
alwismt: Added to  .github/dependabot.yml and  .github/workflows/test.yml

gaby:
@alwismt I ran all the code in the PR through GPT-4 which suggested around 4-5 fixes on each file. One thing that caught my attention is that in all the files it says the implementation is only good for small files, larger files would suffer from performance issues and suggestion using a library to do the minimization. It also mentioned how some tags/etc were not taken into account.

After doing a quick search, reading on reddit found this lib that covers all the use cases: https://github.com/tdewolff/minify

Would it make more sense to use that to build a middleware for fiber rather than try to implement our own minimizing logic?

alwismt:
@gaby I agree that using an external library like  like https://github.com/tdewolff/minify would be a more efficient solution, especially for larger files. I will integrate it into the minify middleware for fiber. Thanks for your suggestion!

gaby:
@alwismt Any progress on this?

gaby:
@alwismt Friendly ping :-)

Issue #594, opened 2023-05-16
fiberi18n error on concurrent requests


tran-the-lam:
Hi team!

I have an issue.  When there are many concurrent requests to our system, fiberi18n throws error like this
```
[2023-05-16T04:55:32.634958995Z] 57398fb9-da68-4ea2-b008-51da27c5a0ab:: 200  -  GET      /threads 7.703849ms -​
fatal error: concurrent map writes
fatal error: concurrent map writes

goroutine 172 [running]:
github.com/gofiber/contrib/fiberi18n.New.func1(0xc0002282c0)
	/go/pkg/mod/github.com/gofiber/contrib/fiberi18n@v0.1.0/i18n.go:42 +0x209
github.com/gofiber/fiber/v2.(*App).next(0xc0001b8480, 0xc0002282c0)
	/go/pkg/mod/github.com/gofiber/fiber/v2@v2.42.0/router.go:134 +0x1b6
github.com/gofiber/fiber/v2.(*App).handler(0xc0001b8480, 0x4cc317?)
	/go/pkg/mod/github.com/gofiber/fiber/v2@v2.42.0/router.go:160 +0x87
github.com/valyala/fasthttp.(*Server).serveConn(0xc00054c200, {0xf7e6f8?, 0xc0000aa1b0})
	/go/pkg/mod/github.com/valyala/fasthttp@v1.45.0/server.go:2371 +0x11d3
github.com/valyala/fasthttp.(*workerPool).workerFunc(0xc000539720, 0xc0000c0b60)
	/go/pkg/mod/github.com/valyala/fasthttp@v1.45.0/workerpool.go:224 +0xa9
github.com/valyala/fasthttp.(*workerPool).getCh.func1()
	/go/pkg/mod/github.com/valyala/fasthttp@v1.45.0/workerpool.go:196 +0x38
created by github.com/valyala/fasthttp.(*workerPool).getCh
	/go/pkg/mod/github.com/valyala/fasthttp@v1.45.0/workerpool.go:195 +0x1b0

goroutine 1 [IO wait, 1 minutes]:
internal/poll.runtime_pollWait(0x7fb9247119b8, 0x72)
	/usr/local/go/src/runtime/netpoll.go:306 +0x89
internal/poll.(*pollDesc).wait(0xc0004ad100?, 0x4?, 0x0)
	/usr/local/go/src/internal/poll/fd_poll_runtime.go:84 +0x32
internal/poll.(*pollDesc).waitRead(...)
[...]

PR #595, opened 2023-05-16
fix fiberi18n error on concurrent requests

Skyenought:
#### What type of PR is this?
fix

#### (Optional) Which issue(s) this PR fixes:

https://github.com/gofiber/contrib/issues/594

#### How to resolve this issue:

In fact the template only needs to be read once when the program is initialised.

But I didn't do that 😰


ReneWerner87:
can you provide more unittests so that next time we can find this problem right away through race detection etc.?

Skyenought:
OK, will update unit test later

tran-the-lam:
@ReneWerner87 @Skyenought Could you merge pr in to main branch?

ReneWerner87 (approved):

PR #596, opened 2023-05-17
:recycle: refactor: merge websocket middleware to contrib repo

efectn:
Same reason as https://github.com/gofiber/fiber/commit/c7c37d9b504f530a23ee1b03df872e8e0f17002f


ReneWerner87:
thx

.github/workflows/test.yml
```diff
@@ -58,3 +58,5 @@ jobs:
                 run: cd ./fiberi18n && go test ./... -v -race
             -   name: Test fiberzerolog Middleware
                 run: cd ./fiberzerolog && go test ./... -v -race
+            -   name: Test websocket Middleware
```
gaby, line 61: Can we add go1.20 to the test matrix here too?

websocket/go.mod
```diff
@@ -0,0 +1,26 @@
+module github.com/gofiber/contrib/websocket
+
+go 1.20
```
gaby, line 3: Should this be go1.20?

gaby (changes requested):

websocket/go.mod
efectn: it's because of i ran go init on go1.20. Has no affect aganist old go versions unless we run go mod tidy with compat

gaby: Ahhhh, ok

gaby (approved):

Issue #632, opened 2023-06-03
🚀 [Feature]: Add ability to ignore health checks in otelfiber


efumagal:
### Feature Description

Something that was done for fiberzap but for otelfiber: https://github.com/gofiber/contrib/pull/272

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

emanuelef:
This one can probably be closed after https://github.com/gofiber/contrib/commit/9cc7ba741b4b40457dc8ddba2309f210417414fe

emanuelef:
Although I guess to be able to use WithNext we need a new tag/release for otelfiber, isn't it ?

ReneWerner87:
yes, with the next release process

PR #633, opened 2023-06-05
OtelFiber: Add Next config setting

emanuelef:
Just an idea for https://github.com/gofiber/contrib/issues/632
If this makes sense I can go on with this idea and start testing it.


ReneWerner87:
Tests and readme adjustments are missing

ReneWerner87:
Why not using the skip middleware or a next function like in all other middlewares

emanuelef:
Thanks for looking into that.
This approach is ok for what I need but let me know if there's a better way.
I looked at otelzap and did something similar.

emanuelef:
@ReneWerner87 just wanted to ask if this feature is relevant, otherwise I can close the PR.
I'm using this in my fork but might be we have other way to do that in otelfiber.
Some other go web frameworks have a similar feature for their OTel instrumentation.

ReneWerner87:
Sorry i lost the focus on this, will check it in next days and then let you know that the decision is

ReneWerner87:
first thank you for the effort
your feature works and offers added value

however, it would be better if we follow the lead of all other middlewares and add a config setting for a check method "Next: func(c *fiber.Ctx) bool"

We have this in all other middlewares and the users should be familiar with it.
https://github.com/search?q=repo%3Agofiber%2Ffiber%20Next%3A&type=code
https://docs.gofiber.io/api/middleware/csrf and so on

this would also give us more flexibility, because what if you don't want to use a url, but a parameter, cookie or header to skip this functionality.

i know in the 2 other contrib middlewares(Fiberzap, otelfiber) a skipurl config property was also added
this one i had unfortunately overlooked and i will add improvement reports/tasks to change this as deprecated and add a next method configuration property there as well

So that this concept is found in all middlewares and the consumers see a red trail

you would help us a lot if you could adjust your pull request accordingly

emanuelef:
@ReneWerner87 Thanks for looking into that. Yes, with Next it would be much better, I'll try to adapt my PR to use that.
Not sure if it should be a separate PR and close this one.

otelfiber/config.go
```diff
@@ -28,6 +29,12 @@ func (o optionFunc) apply(c *config) {
 	o(c)
 }
 
+func Next(f func(ctx *fiber.Ctx) bool) Option {
```
emanuelef, line 32: Not sure if to follow the pattern here should be WithNext, it's a bit different to the standard fibre middleware where we have ConfigDefault

ReneWerner87: name it withNext, its a shortcut function to work with the defaults

gaby (approved):
LGTM

ReneWerner87:
@emanuelef 

- [ ] can you add the config possiblities to the README.md

https://github.com/gofiber/contrib/blob/main/otelfiber/README.md

like the others
https://github.com/gofiber/contrib/blob/main/opafiber/README.md#config
https://github.com/gofiber/contrib/blob/main/paseto/README.md#config
https://github.com/gofiber/contrib/tree/main/fibersentry#config
....
and maybe a example for the usage

we will use these readme's later in our https://docs.gofiber.io/ and there it is important to have all the information about this packet

this should show up as an issue report in the repository itself or discord channel

because only a well documented software can be well used

emanuelef:
I've added two new sections to the README, but I have some questions there.
Most of the example seems to be inspired by another popular go web framework, I was wondering if we want to have something different. Also the tests are almost exactly the same and are not covering all of the possible configs (WithSpanNameFormatter is not tested).

gaby:
@emanuelef Totally agree, that could be done in a separate PR so this one can be merged. :-)

PR #640, opened 2023-06-15
Fix use of multi key in JWT middleware

t0jc:
Fix multiple keys


gaby:
@t0jc is there a test we can add to trigger this bug?

gaby:
The fact that tests didnt fail before/after this change means we need to add a Test Case for it.

t0jc:
Sorry, this is my first time pulling the repository. Should I submit a complete test case or just provide an example?

t0jc:
> The fact that tests didnt fail before/after this change means we need to add a Test Case for it.


I add test for multi key.

gaby (approved):

efectn (approved):

Issue #624, opened 2023-05-27
🐛 [Bug]: Websocket middleware is not compatible with Recover middleware


wyyadd:
### Bug Description

Hi team!
I have an issue. Websocket middleware is not compatible with Recover middleware.  
If panic happens in websocket handler,  process will quit.

Possible reason:
In this code block, websocket middleware reigsters a handler to FastHTTPUpgrader.
``` golang
        if err := upgrader.Upgrade(c.Context(), func(fconn *websocket.Conn) {
            conn.Conn = fconn
            defer releaseConn(conn)
            handler(conn)
        }); err != nil { // Upgrading required
            return fiber.ErrUpgradeRequired
        }
```
FastHTTPUpgrader also registers a **HijackHandler**.
``` golang
func (u *FastHTTPUpgrader) Upgrade(ctx *fasthttp.RequestCtx, handler FastHTTPHandler) error {
        // ....
	ctx.Hijack(func(netConn net.Conn) {
                // ....
		handler(c)
                // ....
	})
        // ....
	return nil
}
```
FastHttp runs hijackConnHandler in a **new goroutine**. If a panic occurs within this goroutine, the Recover middleware is unable to catch it.
``` golang
func (s *Server) serveConn(c net.Conn) (err error) {
        // ....
	if hijackHandler != nil {
                // ....
		go hijackConnHandler(ctx, hjr, c, s, hijackHandler)
                // ....
	}
        // ....
}
```

### How to Reproduce

``` golang
func main() {
	app := fiber.New()
	app.Use(recover.New(recover.Config{EnableStackTrace: true}))

	app.Get("/foo", websocket.New(func(c *websocket.Conn) {
		panic("foo")
	}))

	_ = app.Listen(":3000")
}
```

### Expected Behavior

```
panic: foo
goroutine 20 [running]:
main.main.func1(0x6b8daf?)
        /home/leo/project/main.go:16 +0x27
github.com/gofiber/contrib/websocket.New.func2.5(0x0?)
        /home/leo/go/pkg/mod/github.com/gofiber/contrib/websocket@v1.0.0/websocket.go:130 +0x75
github.com/fasthttp/websocket.(*FastHTTPUpgrader).Upgrade.func1({0x9cf848, 0xc0000961e0})
        /home/leo/go/pkg/mod/github.com/fasthttp/websocket@v1.5.3/server_fasthttp.go:201 +0x1bf
[...]

PR #681, opened 2023-07-15
add recover config & default recover handler & tests

mstrYoda:
fix: #624

As a default behavior I wrote the error as response message. But it can be changed WDYT about it? @efectn


ReneWerner87 (approved):

ReneWerner87:
@mstrYoda pls handle this
image

mstrYoda:
> @mstrYoda pls handle this
> 
> image
> 
> 
> 
> 

I will patch this a few hours later 👍

efectn:
@mstrYoda can you add support for StackTrace like current recover middleware, it might be useful for some cases.

mstrYoda:
> @mstrYoda can you add support for StackTrace like current recover middleware, it might be useful for some cases.

I pushed the changes, added print stack trace to default recover handler. I did not take any config for recover handler, one can just change it at all including stack handler.

websocket/websocket.go
```diff
@@ -5,208 +5,230 @@
...
+	// per message compression (RFC 7692). Setting this value to true does not
+	// guarantee that compression will be supported. Currently only "no context
+	// takeover" modes are supported.
+	EnableCompression bool
+
+	// Recover is a panic handler function that recovers from panics
+	// Default recover function is used when nil and writes error message in a response field `error`
+	// It prints stack trace to the stderr by default
+	// Optional. Default: defaultRecover
+	Recover func(*Conn)
```
efectn, line 64: Should this field be `RecoverHandler`?

websocket/websocket.go
```diff
@@ -56,13 +59,17 @@ type Config struct {
...
 	// Default recover function is used when nil and writes error message in a response field `error`
+	// It prints stack trace to the stderr by default
 	// Optional. Default: defaultRecover
 	Recover func(*Conn)
 }
 
 func defaultRecover(c *Conn) {
 	if err := recover(); err != nil {
-		c.WriteJSON(fiber.Map{"error": err})
+		_, _ = os.Stderr.WriteString(fmt.Sprintf("panic: %v\n%s\n", err, debug.Stack())) //nolint:errcheck // This will never fail
```
efectn, line 69: I think we can add new config field for stacktrace

websocket/websocket.go
ReneWerner87: @mstrYoda pls change this

mstrYoda: Done 👍

efectn (approved):

Issue #704, opened 2023-08-02
🚀 [Feature]: Stringify zerolog latency


raphielscape:
### Feature Description

Currently, zerolog latency uses `Dur`, which outputs raw duration value compared to other logging middlewares in fiber, changing it to String should make it in line with other logging behaviour.

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

PR #705, opened 2023-08-02
fiberzerolog: Use string for latency field

raphielscape:
Fixes https://github.com/gofiber/contrib/issues/704

### DIFF
--- a/fiberzerolog/zerolog_test.go
+++ b/fiberzerolog/zerolog_test.go
@@ -151,12 +151,8 @@ func Test_Latency(t *testing.T) {
 	_ = json.Unmarshal(buf.Bytes(), &logs)
 
 	latencyStr, ok := logs[FieldLatency].(string)
-	if !ok {
-		t.Errorf("Failed to parse latency from logs")
-	}
-	if !strings.Contains(latencyStr, "ms") {
-		t.Errorf("Latency does not contain 'ms': %s", latencyStr)
-	}
+	utils.AssertEqual(t, true, ok)
+	utils.AssertEqual(t, true, strings.Contains(latencyStr, "ms"))
 	utils.AssertEqual(t, float64(200), logs[FieldStatus])
 }
 
@@ -235,12 +231,8 @@ func Test_Logger_All(t *testing.T) {
 	}
 
 	latencyStr, ok := logs[FieldLatency].(string)
-	if !ok {
-		t.Errorf("Failed to parse latency from logs")
-	}
-	if !strings.Contains(latencyStr, "ms") {
-		t.Errorf("Latency does not contain 'ms': %s", latencyStr)
-	}
+	utils.AssertEqual(t, true, ok)
+	utils.AssertEqual(t, true, strings.Contains(latencyStr, "ms"))
 }
 
 func Test_Response_Body(t *testing.T) {
--- a/fiberzerolog/zerolog_test.go
+++ b/fiberzerolog/zerolog_test.go
@@ -8,7 +8,9 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/fiber/v2/utils"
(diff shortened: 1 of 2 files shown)


ReneWerner87:
Pls provide a documentation change and unittests for what this change involves so that it remains audited and documented

gaby:
@ReneWerner87 The current tests dont check any of the fields used by this middleware

gaby:
Actually it does check fields, but doesnt check latency at all. We can at least add a test to check the field type.

The regular logger and fiberzap all return "Strings" for latency but this middleware is not. The changes are more of a consistency thing.

gaby:
`govulncheck` failure is related to go1.20.7 being released. Related: https://github.com/actions/setup-go/issues/405

raphielscape:
I have added the unit test for latency

gaby:
@raphielscape Thanks! Could you also added in the test at the bottom for "Logger All", the one with all the fields?

gaby:
This test: https://github.com/gofiber/contrib/blob/6fa822cca2e7421e5a844746068ea4986f2f8d5c/fiberzerolog/zerolog_test.go#L143

raphielscape:
Done

gaby:
@raphielscape What I was thinking is adding the latency field on this list here: https://github.com/gofiber/contrib/blob/6fa822cca2e7421e5a844746068ea4986f2f8d5c/fiberzerolog/zerolog_test.go#L152 and then checking the expected value below. 

You added a whole new Fiber App in that test which already has a FiberApp

raphielscape:
> @raphielscape What I was thinking is adding the latency field on this list here:
> 
> https://github.com/gofiber/contrib/blob/6fa822cca2e7421e5a844746068ea4986f2f8d5c/fiberzerolog/zerolog_test.go#L152
> 
> and then checking the expected value below.
> You added a whole new Fiber App in that test which already has a FiberApp

I have explained it in the commit description, we can't use the same app because we can't do Equal Assertion as latency varies between runs

raphielscape:
We can make the first fiber app have a endpoint that returns OK and use that in test instead, though

gaby (approved):
LGTM

gaby:
> @gaby good?

Thanks! That looks good

fiberzerolog/zerolog_test.go
```diff
@@ -181,15 +222,25 @@ func Test_Logger_All(t *testing.T) {
...
 	_ = json.Unmarshal(buf.Bytes(), &logs)
 
-	utils.AssertEqual(t, expected, logs)
+	for key, value := range expected {
+		utils.AssertEqual(t, value, logs[key])
+	}
+
+	latencyStr, ok := logs[FieldLatency].(string)
+	if !ok {
+		t.Errorf("Failed to parse latency from logs")
```
efectn, line 239: use utils.AssertEqual instead

fiberzerolog/zerolog_test.go
```diff
@@ -181,15 +222,25 @@ func Test_Logger_All(t *testing.T) {
...
+	}
+	if !strings.Contains(latencyStr, "ms") {
+		t.Errorf("Latency does not contain 'ms': %s", latencyStr)
```
efectn, line 242: same

fiberzerolog/zerolog_test.go
```diff
@@ -125,6 +127,39 @@ func Test_Logger(t *testing.T) {
...
+	resp, err := app.Test(httptest.NewRequest("GET", "/", nil))
+	utils.AssertEqual(t, nil, err)
+	utils.AssertEqual(t, fiber.StatusOK, resp.StatusCode)
+
+	var logs map[string]any
+	_ = json.Unmarshal(buf.Bytes(), &logs)
+
+	latencyStr, ok := logs[FieldLatency].(string)
+	if !ok {
+		t.Errorf("Failed to parse latency from logs")
```
efectn, line 155: here too

fiberzerolog/zerolog_test.go
raphielscape: We can't use AssertEqual because latency always varies and we can't have an equal value for AssertEqual

gaby: @efectn We can only check the type for this field.

raphielscape:
For the reason why I don't use AssertEqual is that I have no idea how to make an equally valued latency and latency return always varies between test runs, please let me know if there was a way to make the latency in an equilibrium so that we got an equal value that can be compared so that I can use AssertEqual

fiberzerolog/zerolog_test.go
efectn: > @efectn We can only check the type for this field.

Isn't ok boolean? We can check it's true

raphielscape: OK came from interface type assertion, not the latency return itself, we can't AssertEqual that

efectn: > OK came from interface type assertion, not the latency return itself, we can't AssertEqual that

I know, i just mean something like
```go
utils.AssertEqual(t, true, ok)
utils.AssertEqual(t, true, strings.Contains(latencyStr, "ms"))
```

raphielscape: > > OK came from interface type assertion, not the latency return itself, we can't AssertEqual that
> 
> I know, i just mean something like
> 
> ```go
> utils.AssertEqual(t, true, ok)
> utils.AssertEqual(t, true, strings.Contains(latencyStr, "ms"))
> ```

Okay, that works

efectn (approved):

Issue #750, opened 2023-08-31
🚀 [Feature]: fiberzap accepts custom fields


mirusky:
### Feature Description

I'd like to have a custom field that are user defined zap.Fields. 

- For custom structured logging
- To add user defined fields
- Support different entry formats, etc.

### Additional Context (optional)

Currently I have an client that want's http headers response into the logging file, that's post processed later. They want all headers, now the fiberzap doesn't have an option for it. 

### Code Snippet (optional)

```go
package main

import (
	"log"

	"github.com/gofiber/contrib/fiberzap/v2"
	"github.com/gofiber/fiber/v2"
	"go.uber.org/zap"
)

func main() {
	app := fiber.New()
	logger, _ := zap.NewProduction()

	app.Use(fiberzap.New(fiberzap.Config{
		Logger: logger,
		Fields: []string{"custom"},
		CustomFieldFunc: func(c *fiber.Ctx) []zap.Field {
			var fields []zap.Field

			headers := make(map[string]interface{})

			c.Response().Header.VisitAll(func(k, v []byte) {
				headers[string(k)] = string(v)
			})

			fields = append(fields, zap.Any("headers", headers))

			return fields
		},
	}))

	app.Get("/", func(c *fiber.Ctx) error {
		return c.SendString("Hello, World!")
	})

	log.Fatal(app.Listen(":3000"))
}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my suggestion prior to opening this one.
- [X] I understand that improperly formatted feature requests may be closed without explanation.

Jamess-Lucass:
@mirusky Hi, I have literally just ran into a use case for this, have you started working on the feature at all? I don't want to intrude but I have written the code for it before I saw this open issue.

Jamess-Lucass:
My use was using this to inject trace ids into the log, so I can link together APM traces in kibana to specific log records eg.

```go
FieldsFunc: func(c *fiber.Ctx) []zap.Field {
      tr := apm.TransactionFromContext(c.Context())
      traceId := tr.TraceContext().Trace.String()
      
      return []zap.Field{zap.String("trace.id", traceId)}
},
```

mirusky:
@Jamess-Lucass I didn't started at all, busy week. 

[...]

PR #753, opened 2023-09-04
feat(fiberzap): Added 'FieldsFunc' to provide custom log fields

Jamess-Lucass:
Solves #750

### DIFF
--- a/fiberzap/zap.go
+++ b/fiberzap/zap.go
@@ -110,12 +110,13 @@ func New(config ...Config) fiber.Handler {
 
 		// Add fields
 		fields := make([]zap.Field, 0, len(cfg.Fields)+1)
+		fields = append(fields, zap.Error(err))
 
 		if cfg.FieldsFunc != nil {
-			fields = cfg.FieldsFunc(c)
+			for _, field := range cfg.FieldsFunc(c) {
+				fields = append(fields, field)
+			}
 		}
-
-		fields = append(fields, zap.Error(err))
 
 		for _, field := range cfg.Fields {
 			switch field {
--- a/fiberzap/zap.go
+++ b/fiberzap/zap.go
@@ -110,6 +110,11 @@ func New(config ...Config) fiber.Handler {
 
 		// Add fields
 		fields := make([]zap.Field, 0, len(cfg.Fields)+1)
+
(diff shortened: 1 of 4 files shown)


fiberzap/zap_test.go
```diff
@@ -441,3 +441,26 @@ func Test_LoggerLevelsAndMessagesSingle(t *testing.T) {
...
+
+	app.Get("/", func(c *fiber.Ctx) error {
+		return c.SendString("hello")
+	})
+
+	resp, err := app.Test(httptest.NewRequest("GET", "/", nil))
+
+	utils.AssertEqual(t, nil, err)
+	utils.AssertEqual(t, fiber.StatusOK, resp.StatusCode)
+	utils.AssertEqual(t, "test", logs.All()[0].Context[0].String)
```
gaby, line 465: Can you add something like this to verify all the other expected fields are preset:

```go
expected := map[string]interface{}{
		"body":          "",
		"ip":            "0.0.0.0",
		"host":          "example.com",
		"url":           "/?foo=bar",
		"method":        "GET",
		"route":         "/",
		"protocol":      "http",
		"pid":           strconv.Itoa(os.Getpid()),
		"queryParams":   "foo=bar",
		"resBody":       "Cannot GET /",
		"bytesReceived": int64(0),
		"bytesSent":     int64(12),
	}
	utils.AssertEqual(t, expected, logs.All()[0].ContextMap())
```

gaby (changes requested):

fiberzap/zap_test.go
Jamess-Lucass: See latest commit, took inspiration from `Test_Logger_All`

gaby (approved):

fiberzap/zap.go
```diff
@@ -110,6 +110,11 @@ func New(config ...Config) fiber.Handler {
 
 		// Add fields
 		fields := make([]zap.Field, 0, len(cfg.Fields)+1)
+
+		if cfg.FieldsFunc != nil {
+			fields = cfg.FieldsFunc(c)
```
efectn, line 115: Appending would be better

Jamess-Lucass:
Thanks everyone <3

PR #813, opened 2023-10-13
FiberZerolog: Add user context

dotX12:
Fix: It is impossible to use hooks that take something out of the context and write a log (example: span_id, trace_id)


dotX12:
Fix: It is impossible to use hooks that take something out of the context and write a log (example: span_id, trace_id)

gaby (approved):

efectn (approved):

PR #900, opened 2024-01-05
Update documentation on contrib/casbin

complib2891:
to add notes on Config's Enforcer setting

### DIFF
--- a/casbin/README.md
+++ b/casbin/README.md
@@ -153,3 +153,63 @@ func main() {
   app.Listen(":8080")
 }
 ```
+## Notes on Config's Enforcer
+
+Use this property instead, if you
+- manage to test out above example with success
+- don't want to expose model and policy to filesystem, thus reveal less and prevent filesystem modifications.
+- original casbin.NewEnforcer() supports way more [adapters](https://casbin.org/docs/adapters) and more ways to initialize *casbin.Enforcer 
+
+below example doesn't require `./model.conf` and `./policy.csv` to be present.
+```go
+package main
+
+import (
+  ...
+  cb "github.com/casbin/casbin/v2"
+  "github.com/casbin/casbin/v2/model"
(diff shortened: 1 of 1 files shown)


t101804 (approved):

casbin/README.md
```diff
@@ -153,3 +153,63 @@ func main() {
   app.Listen(":8080")
 }
 ```
+## Notes on Config's Enforcer
+
+Use this property instead, if you
```
gaby, line 158: The wording here needs some work.

casbin/README.md
```diff
@@ -153,3 +153,63 @@ func main() {
...
+- manage to test out above example with success
```
gaby, line 159: I'm assuming the purpose is to be able to have policies in memory instead of a file?

complib2891: yeah, that's the idea, some instance where the policy is static instead of dynamic, perhaps it's better to hardcode the policy in memory instead of filesystem, at least that's what I'm thinking. because casbin can be used as static or dynamic policy i presumed, if anyone set the filesystem mode too widely then anyone could modify the policy, and maybe wait for the system to crash and restart, then the modified policy will get read.

disclosure, I'm not a security expert, just figure it's safer to just load these in memory instead of files since developer should understand which func call should be protected or not.

casbin/README.md
complib2891: you're free to modify to whatever you see fit. I'm just sharing what I've studied so far.

casbin/README.md
complib2891: the example I provided, doesn't use any Auto-Save feature because json-adapter (by casbin folks) is static. for those dynamic adapters with Auto-Save adapter, it's probably for those software system that allows their admin to modify the policy and persist in database(s)

PR #932, opened 2024-01-29
fix: when body is stream middleware read everything into memory first, no streaming to client

Alantoo:
Bug Description
When application stream some file middleware reads all the streamed data into the memory to calculate response size which can consume all the available memory.

How to Reproduce
- Stream some binary data from handler

Expected Behavior
The middleware should skip response/request body size calculation when body is stream

### DIFF
--- a/otelfiber/fiber.go
+++ b/otelfiber/fiber.go
@@ -96,7 +96,8 @@ func Middleware(opts ...Option) fiber.Handler {
 		copy(responseMetricAttrs, requestMetricsAttrs)
 
 		reqHeader := make(http.Header)
-		c.Request().Header.VisitAll(func(k, v []byte) {
+		request := c.Request()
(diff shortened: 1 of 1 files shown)


ReneWerner87:
@Alantoo could you pls provide a unittest for this case?

ReneWerner87:
@Alantoo friendly ping

Alantoo:
> @Alantoo friendly ping

Hey. Sorry for long delay. Unfortunately had no time to work on it. I've pushed test to make sure response size computed in lazy way correctly. Please let me know if its fine or I should fix anything else.
Thank you

ReneWerner87:
@Alantoo pls fix this
image

ReneWerner87:
@Alantoo 
image


but we still support go 1.19.x

PR #965, opened 2024-02-21
Improve Dict log for reqHeaders & resHeaders

roman-wb:
Hi, @ReneWerner87!

What do you think about it?

1) Improved request headers log:
Was:
`{"method":"POST"," header_key":"header_value"} `
Became with sub dict:
`{"method":"POST", "reqHeaders": {"header_key":"header_value"}}
`
2) Added response headers log (with field FieldResHeaders = "resHeaders"):
`{"method":"POST", "resHeaders": {"header_key":"header_value"}}`

### DIFF
--- a/fiberzerolog/config.go
+++ b/fiberzerolog/config.go
@@ -31,6 +31,7 @@ const (
 	FieldRequestID     = "requestId"
 	FieldError         = "error"
 	FieldReqHeaders    = "reqHeaders"
+	FieldResHeaders    = "resHeaders"
 )
 
 // Config defines the config for middleware.
@@ -78,6 +79,13 @@ type Config struct {
 	//
 	// Optional. Default: {"ip", "latency", "status", "method", "url", "error"}
 	Fields []string
+
+	// Wrap headers to dictionary.
+	// If false: {"method":"POST", "header-key":"header value"}
+	// If true: {"method":"POST", "reqHeaders": {"header-key":"header value"}}
+	//
+	// Optional. Default: false
+	WrapHeaders bool
 
 	// Custom response messages.
 	// Response codes >= 500 will be logged with Messages[0].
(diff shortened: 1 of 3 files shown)


ReneWerner87:
thx for the PR, but iam not really an expert for this MW 
i will tag someone with more exp

ReneWerner87:
@iredmail @raphielscape @dotX12 what do you think?

raphielscape:
Seems like a breaking change for me, but overall it's fine

dotX12:
This will affect users who have header filters configured, for example, dashboards in Grafana, i.e. this is a serious change.

But in the long run it looks sensible, I’m all for it.
LGTM...

ReneWerner87:
so it's a breaking change
would it be possible to design this change without breaking?

dotX12:
> so it's a breaking change would it be possible to design this change without breaking?

I would add a flag so that users can choose the format for listing headers.
Something like:
```
HeadersFormat: rolled | unrolled
```
maybe
```
HeadersFormat : flat | object
```
or
```
WrapHeaders: true / false
```
In this case, we won't break anything :)

roman-wb:
Thx, without breaking done)

fiberzerolog/config.go
```diff
@@ -79,6 +80,13 @@ type Config struct {
 	// Optional. Default: {"ip", "latency", "status", "method", "url", "error"}
 	Fields []string
 
+	// Wrap headers to dictionary.
+	// If false: {"method":"POST", "header-key":"header value"}
+	// If true: {"method":"POST", "reqHeaders": {"header-key":"header value"}}
+	//
+	// Optional. Default: false
+	WrapHeaders bool
```
ReneWerner87, line 88: @roman-wb 
documentation is important, pls add some infos to the markdown file
add it to the table 
https://docs.gofiber.io/contrib/next/fiberzerolog/

roman-wb: Added to this repo and Docs https://github.com/gofiber/docs/pull/388

ReneWerner87: Thanks, only the adaptation of the readme in this repository is necessary
The docs repository is automatically synchronized on release

ReneWerner87 (approved):

PR #980, opened 2024-02-22
fix(#979): GetHubFromContext may crash instead of returning nil

### LINKED ISSUE #979: 🐛 [Bug]: FiberSentry: GetHubFromContext() never returns nil
### Bug Description

When using the sentry function fibersentry/sentry.go:GetHubFromContext() there is a runtime panic when the appropriate middleware was not registered. I understand, that one should not call GetHubFromContext() if there is no appropriate middleware registered, but it doesn't make sense to crash the entire app, because of a faulty interface conversion.

Also when following the readme example, there is an explicit nil check, suggesting that in case of a faulty implementation nil is being returned (which is not the case).

### How to Reproduce

1. Initialize Fiber App
2. Do not add sentry middleware or initialization
3. Call GetHubFromContext()
4. Panic

### Expected Behavior

Instead of crashing nil is being returned so that the developer can handle the error appropriately.

### Contrib package Version

 main branch

### Code Snippet (optional)

```go
package main

import (
	"github.com/gofiber/contrib/fibersentry"
	"github.com/gofiber/fiber/v2"
)

func main() {
	app := fiber.New()

	// Intentionally not adding the sentry middleware

	app.Get("/", func(c *fiber.Ctx) error {
		hub := fibersentry.GetHubFromContext(c)
		if hub == nil {
			// Unreachable: the call above panics
		}

		return c.SendStatus(fiber.StatusOK)
	})

	app.Listen(":3000")
}
```

### Checklist:

- [X] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

kohnalex:
fix #979

## Summary by CodeRabbit

- **Refactor**
	- Improved error handling and clarity in hub retrieval from context by splitting the functionality into two distinct methods.
- **Documentation**
	- Updated `fibersentry` package README with new usage instructions emphasizing `*sentry.Hub` instance access methods.
- **Tests**
	- Restructured test cases in `sentry_test.go` to cover various middleware states for handling panics and capturing events.


ReneWerner87 (approved):

kohnalex:
Thanks for approving! I habe no write access, so a merge would be nice.

fibersentry/sentry.go
```diff
@@ -54,5 +53,9 @@ func New(config ...Config) fiber.Handler {
 }
 
 func GetHubFromContext(ctx *fiber.Ctx) *sentry.Hub {
```
efectn, line 55: Can we add new method MustGetFromContext that panics if can't cast type?

ReneWerner87:
@kohnalex pls add description to the readme

gaby:
@kohnalex Can you update the README with the new function. Also add a unit-test for it. Thanks!

ReneWerner87:
@kohnalex friendly ping, pls check the last hints

- [ ] Add description in readme 
- [ ] Add unittests for the new/adjusted code

kohnalex:
Sorry for the wait... I had to restructure the test cases a little so `app.Use(New())` will be called after the test cases I have added.

xEricL:
What else is needed to merge this?

ReneWerner87 (approved):

PR #995, opened 2024-02-25
Option using fields in snake_case.

roman-wb:
Hi!

Added option FieldsSnakeCase:bool
Use snake case for fields: FieldResBody, FieldQueryParams, FieldBytesReceived, FieldBytesSent, FieldRequestId, FieldReqHeaders, FieldResHeaders.

If false (default): `{"method":"POST", "resBody":"v", "queryParams":"v"}`
If true: `{"method":"POST", "res_body":"v", "query_params":"v"}`

## Summary by CodeRabbit

- **New Features**
	- Introduced new logging fields including response body, query parameters, bytes received/sent, request ID, and request/response headers.
	- Added a configuration option to toggle field naming convention between snake case and default.


ReneWerner87 (changes requested):
why is this feature necessary ?
we can't support every spelling, if another spelling should be enabled, we have to add another much more generic way to not extend the configuration for each spelling to customize it

roman-wb:
> why is this feature necessary ? we can't support every spelling, if another spelling should be enabled, we have to add another much more generic way to not extend the configuration for each spelling to customize it

Often JSON keys in text presentation used as snake case.

As variant 2: adding new keys (example FieldResBodySC etc) and developer will set in option Fields.

sixcolors:
@ReneWerner87 snake case is pretty common is json; eg https://developer.apple.com/documentation/foundation/jsondecoder/keydecodingstrategy/convertfromsnakecase#

ReneWerner87 (approved):

ReneWerner87:
Ok valid

ReneWerner87:
Will merge it tomorrow

ReneWerner87 (approved):

PR #1085, opened 2024-04-20
Swagger: Add support for File Content

willfaught:
I need to be able to embed a swagger.json file into a Go program using the embed package. Config.FileContent enables that.

## Summary by CodeRabbit

- **New Features**
	- Enhanced API documentation capabilities by allowing Swagger file content to be provided directly, in addition to the traditional file path method.
- **Tests**
	- Added comprehensive tests for new Swagger file handling configurations and multiple API documentation instances.


willfaught:
Synced my fork. Rebased on main.

swagger/swagger.go
```diff
@@ -85,16 +91,20 @@ func New(config ...Config) fiber.Handler {
 		}
 	}
 
-	// Verify Swagger file exists
-	if _, err := os.Stat(cfg.FilePath); os.IsNotExist(err) {
-		panic(fmt.Errorf("%s file does not exist", cfg.FilePath))
-	}
+	rawSpec := cfg.FileContent
```
gaby, line 94: This needs to be inside the IF above where cfg is override

swagger/swagger_test.go
```diff
@@ -259,3 +266,229 @@ func TestNew(t *testing.T) {
 		require.Equal(t, "success", string(bodyBytes))
 	})
 }
+
+func TestNewWithFileContent(t *testing.T) {
```
gaby, line 270: This is passing but not using FileContent, we need an different swagger.json file to verify the data we get back from `/swagger` matches the content.

gaby (changes requested):
Overall looks good, reorder the New() and add swagger.json to be used for FileContent testing that's different than the current one. 

For example copy the current swagger.json, change the version # and make a new file. Use this file in the test verify the file returned has that version instead of the current one.

sixcolors:
@willfaught looks open to me....

gaby:
> @willfaught looks open to me....

Different PR, the previous one was closed

swagger/swagger.go
willfaught: I don't follow. rawSpec is the data to use. If FileContent is set, then we use that, otherwise we load it from FilePath in the following if-statement.

While looking at the above if-statement, I realized that cfg.FilePath should only be set from DefaultConfig if cfg.FileContent is also empty, so I added that in [75a50b6](https://github.com/gofiber/contrib/pull/1085/commits/75a50b64281d49fc38f51810749967444941eb44).

willfaught: >While looking at the above if-statement, I realized that cfg.FilePath should only be set from DefaultConfig if cfg.FileContent is also empty, so I added that in [75a50b6](https://github.com/gofiber/contrib/pull/1085/commits/75a50b64281d49fc38f51810749967444941eb44).

This was wrong. cfg.FilePath still needs to be set like before for serving /swagger.json. Force pushed to remove the fixup commit.

gaby: Hmmm, it wouldnt use FileContent then. My suspicion is that the tests are using the file not the content since they both reference "swagger.json". We need a new file to compare properly

swagger/swagger_test.go
willfaught: >This is passing but not using FileContent

I don't follow. Almost every subtest sets FileContent. For example:

```go
func TestNewWithFileContent(t *testing.T) {
	t.Run("Endpoint check with only custom path", func(t *testing.T) {
		app := fiber.New()

		cfg := Config{
			Path:        "custompath",
			FileContent: swaggerJSON,
		}
```

>we need an different swagger.json file to verify the data we get back from /swagger matches the content.

Why does TestNewWithFileContent need to do that, but not TestNew? Also, why would we need a different file for that?

gaby: @willfaught Lets try something, in the `NewWithFileContent` tests, set FilePath to `myswagger.json`. Internally this should expose the spec on `/myswagger.json`, but use the embed swagger.json as the content.

If that passes we should be good to go

swagger/swagger.go
willfaught: >Hmmm, it wouldnt use FileContent then.

What do you mean? It always uses FileContent if its length is > 0. Here is the full context:

```go
	rawSpec := cfg.FileContent
	if len(rawSpec) == 0 {
		// Verify Swagger file exists
		_, err := os.Stat(cfg.FilePath)
		if os.IsNotExist(err) {
			panic(fmt.Errorf("%s file does not exist", cfg.FilePath))
		}

		// Read Swagger Spec into memory
		rawSpec, err = os.ReadFile(cfg.FilePath)
		if err != nil {
			log.Fatalf("Failed to read provided Swagger file (%s): %v", cfg.FilePath, err.Error())
			panic(err)
		}
	}
```

>My suspicion is that the tests are using the file not the content since they both reference "swagger.json". We need a new file to compare properly

I'll change the FilePath in the tests to avoid the potential overlap.

willfaught: [ac66002](https://github.com/gofiber/contrib/pull/1085/commits/ac66002a48f4eddedd2e70129e1dbba524d04196)

swagger/swagger_test.go
willfaught: [ac66002](https://github.com/gofiber/contrib/pull/1085/commits/ac66002a48f4eddedd2e70129e1dbba524d04196)

gaby: Awesome, it passed

gaby (approved):
LGTM

gaby:
Ping @gofiber/maintainers

ReneWerner87 (approved):

swagger/swagger.go
```diff
@@ -32,6 +32,12 @@ type Config struct {
 	// Optional. Default: ./swagger.json
 	FilePath string
 
+	// FileContent for the content of the swagger.json or swagger.yaml file.
+	// If provided, FilePath will not be read.
+	//
+	// Optional. Default: nil
+	FileContent []byte
```
ReneWerner87, line 39: pls add something in the readme section for this new setting

ReneWerner87 (changes requested):
- [ ] documentation missing

ReneWerner87:
@willfaught can you please add a description for the new feature in the markdown files and maybe an example
otherwise hardly anyone will discover and use the feature

willfaught:
@ReneWerner87 [f0070f2](https://github.com/gofiber/contrib/pull/1085/commits/f0070f2b59135c32d77b7d1177d77ba687e3d40f)

swagger/swagger.go
willfaught: f0070f2

swagger/README.md
```diff
@@ -63,6 +63,19 @@ cfg := swagger.Config{
...
+cfg := swagger.Config{
+    BasePath:    "/",
+    FilePath:    "./docs/swagger.json",
+    FileContent: mySwaggerByteSlice,
+    Path:        "swagger",
+    Title:       "Swagger API Docs",
+}
+
+app.Use(swagger.New(cfg))
+```
```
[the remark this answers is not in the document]
ReneWerner87: no, the current is enough

ReneWerner87 (approved):

PR #1202, opened 2024-11-21
Abstract Fiber Context with Interface for Version Independence

ReneWerner87:
This PR introduces an abstraction for the Fiber context to make Fiber contrib libraries independent of the Fiber version being used.

Key Changes:

1.	Context Interface:
	- Created a Context interface that defines the required methods from fiber.Ctx for use in the middleware.
	- This allows the middleware to work with any implementation of Context, decoupling it from a specific Fiber version.
2.	Generics Support:
	- Introduced generics in the middleware handler with Handler[T Context] to ensure type safety and support for different context implementations.
3.	Test Adjustments:
	- Updated all middleware tests to use the new generic-based handler and fiber.Ctx implementation.

Known Issue:

Currently, the fiber.Ctx interface presents a limitation for complete version independence. Methods such as Status() return fiber.Ctx itself, which tightly couples the interface to the original Fiber implementation. This behavior makes the abstraction less flexible and middleware dependent on a specific Fiber version.
A fix for this issue is planned by modifying the Context interface to return the abstracted Context instead of fiber.Ctx.

This PR is an important step towards making Fiber contrib libraries more flexible, reusable, and compatible across different Fiber versions. Further refinements will follow to address the self-referential return issue and fully decouple the middleware from the Fiber framework.

### DIFF
--- a/monitor/config.go
+++ b/monitor/config.go
@@ -2,8 +2,6 @@ package monitor
 
 import (
 	"time"
-
-	"github.com/gofiber/fiber/v3"
 )
 
 // Config defines the config for middleware.
@@ -26,7 +24,7 @@ type Config struct {
 	// Next defines a function to skip this middleware when returned true.
 	//
 	// Optional. Default: nil
-	Next func(c *fiber.Ctx) bool
+	Next func(c Context) bool
 
 	// Custom HTML Code to Head Section(Before End)
 	//
--- a/monitor/context.go
+++ b/monitor/context.go
@@ -1,0 +1,18 @@
+package monitor
+
+import "github.com/gofiber/fiber/v3"
+
+// Context is an interface that abstracts the Fiber context.
+type Context interface {
+	Method(override ...string) string
+	Get(key string, defaultValue ...string) string
+	// TODO: return self is a problem
+	Status(status int) fiber.Ctx
+	JSON(data any, ctype ...string) error
+	Next() error
+	Set(key, val string)
+	SendString(body string) error
+}
+
+// Handler is a function type that takes a Context and returns an error.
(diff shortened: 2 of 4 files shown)


ReneWerner87:
@coderabbitai what would be your suggestion to solve the existing problem and solve the idea of the independent version

theoretically, other middlewares could also use other functions, which also give themselves back again

ReneWerner87:
@coderabbitai but using the fiber.Ctx would mean that the code is not independent of the version used

ReneWerner87:
@coderabbitai we only want to support v3 and higher versions as v3 added the ctx interface for the first time

ReneWerner87:
@coderabbitai create a new issue in the fiber repository

all information should be contained there

we have decided to implement the fluent concept for the methods 'status' and 'type' as shown in the following articles
https://appliedgo.com/blog/generic-interface-functions
https://gotipplay.golang.org/p/71m5yxV7ktd
https://medium.com/@mier85/self-referencing-interfaces-in-golang-1-18-bcd6b5701992

there, a generic is used for the interface, which is then used as a return for the methods

in other languages this is achieved with the keyword `self`, but golang will not add this because it means too many changes and there is a possibility with the generics

the issue for the change should therefore deal with the change to the interface

also noticed that if you register a custom context and you use these fluent methods, you would switch the class
to prevent this, you should also have the generic interface for the ctx on the app, which should also be implemented

to make it easier to use, there should be several `New` methods later, so that people who do not want to customize the context have no change
i.e. the normal context is used in the fiber.new method and the custom context is used in another one

ReneWerner87:
@coderabbitai you are welcome to create the issue in the fiber repository

PR #1239, opened 2025-03-22
chore: upgrade Fiber framework from v2 to v3

izetmolla:
- Update Fiber dependency in go.mod and related files
- Adjust code to accommodate breaking changes in Fiber v3 API
- Verify JWT functionality remains intact after the upgrade


gaby:
@izetmolla v3 is still under development. We will upgrade these packages once close to release.

PR #1287, opened 2025-04-26
websocket compatibility with fiber v3

andressg79:


gaby:
@andressg79 We are doing these migrations once V3 is closer to release

PR #1288, opened 2025-04-29
feat(websocket): update to support fiber V3 beta.4

liwuv587:
update websocket to support fiber beta.4


ReneWerner87:
we already updated the code in the v3-beta branch with our migration tool

thx for you effort

PR #1305, opened 2025-05-14
JWT: Fix usage with custom headers.

martinei:
A recent commit broke support for custom headers via `TokenLookup: "header:X-Token"` by enforcing a whitespace "between" scheme and value. Since custom Headers don't have a schema this enforce a space at the beginning of custom header values, which is not possible.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved JWT token extraction from custom HTTP headers, ensuring correct authentication even when no authentication scheme is specified.
- **Tests**
  - Added tests to verify JWT authentication works with tokens provided in custom headers.

### DIFF
--- a/jwt/jwt_test.go
+++ b/jwt/jwt_test.go
@@ -184,6 +184,35 @@ func TestJwtFromHeader(t *testing.T) {
 		}
 	})
 
+	t.Run("custom", func(t *testing.T) {
+		for _, test := range hamac {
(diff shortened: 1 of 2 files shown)


ReneWerner87 (approved):

ReneWerner87:
bug was introduced in https://github.com/gofiber/contrib/pull/1190/files

PR #1909, opened 2026-05-10
🐛 fix(socketio): stop processing polling batch after auth rejection disconnect

gaby:
### Motivation
- Prevent a rejected polling `CONNECT` (e.g. via the deferred `New` callback) from allowing subsequent record-separated packets in the same HTTP POST to be dispatched to application listeners.

### Description
- After `kws.dispatchEIOPacket(packet)` in `ingestPolling` (`v3/socketio/polling.go`), the code now checks `kws.IsAlive()` and `break`s the batch loop if the connection is no longer alive so later packets in the same POST are not processed.

### DIFF
--- a/v3/socketio/polling.go
+++ b/v3/socketio/polling.go
@@ -582,6 +582,9 @@ func ingestPolling(c fiber.Ctx, kws *Websocket) error {
 		// additional per-packet allocation.
 		kws.dispatchEIOPacket(packet)
 		count++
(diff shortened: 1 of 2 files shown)


gaby:
@copilot+gpt-5.4 apply changes based on the comments in [this thread](https://github.com/gofiber/contrib/pull/1909#pullrequestreview-4259761281)

ReneWerner87 (approved):

PR #1922, opened 2026-05-13
🐛 bug(circuitbreaker): prevent default OnClose from advancing handler chain

gaby:
### Motivation
- Fix a security regression where the default `OnClose` called `c.Next()` and could resume the Fiber handler chain a second time during a half-open → closed transition, enabling auth/authz bypass when the circuitbreaker is mounted before security middleware.
- Preserve existing default failure semantics while eliminating the unsafe default chain-resume behavior.

### Description
- Change default `OnClose` in `v3/circuitbreaker/circuitbreaker.go` from `return c.Next()` to a no-op `return nil` to avoid re-entering the handler chain after a half-open success.
- Leave `IsFailure` and other default behaviors unchanged so only the unsafe chain-resume is removed.
- Add a regression test `TestDefaultOnCloseDoesNotAdvanceChainTwice` in `v3/circuitbreaker/circuitbreaker_test.go` that mounts the middleware before an auth-like middleware and verifies a 401 does not result in the protected handler being invoked while still allowing the circuit to close.

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted circuit breaker behavior to prevent duplicate request processing when transitioning between states.

* **Tests**
  * Added test coverage to verify proper request handling during circuit state transitions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gofiber/contrib/pull/1922)


gaby:
@copilot+gpt-5.4 Address all the review comments

gaby:
@copilot+gpt-5.4 add go 1.26.x to the test suit for circuitbreaker

ReneWerner87 (approved):

PR #1939, opened 2026-05-22
feat: add Skip check after request handling to support downstream middleware context

piyongcai-liucai:
## Description
This PR changes the order of the `Next` skip logic in the Zerolog middleware to allow the skip decision to depend on data set by downstream middlewares or handlers.

Previously the middleware checked `cfg.Next` **before** executing the rest of the handler chain (`c.Next()`). This forced the logger middleware to be placed **first** in the middleware stack if you wanted to access any context values (e.g., user roles, request IDs) inside the skip function. This ordering constraint is unnatural and limits composability.

Now the request is fully processed **before** evaluating `cfg.Next`, making it possible to skip logging based on the final request context, including values set by any middleware that runs after the logger.

## Motivation
- It's common to skip logging for certain endpoints or user roles. The old behaviour forced the logger to be registered before authentication/authorization middlewares, which is counterintuitive.
- Downstream middleware often enriches the context with information needed for logging decisions (e.g., user ID, tenant ID). Moving the check allows a cleaner, more flexible setup.
- This change brings the behaviour closer to other logging middlewares in the Fiber ecosystem.

### Before
```go
if cfg.Next != nil && cfg.Next(c) {
    return c.Next() // returns result of c.Next(), which re-executes the handler chain
}

start := time.Now()

chainErr := c.Next()
if chainErr != nil {
    if err := c.App().ErrorHandler(c, chainErr); err != nil {
        _ = c.SendStatus(fiber.StatusInternalServerError)
    }
}

// ... logging ...
```

### After
```go
start := time.Now()

chainErr := c.Next()
if chainErr != nil {
    if err := c.App().ErrorHandler(c, chainErr); err != nil {
        _ = c.SendStatus(fiber.StatusInternalServerError)
    }
}

if cfg.Next != nil && cfg.Next(c) {
    return nil // request already processed, no error to propagate
}

// ... logging ...
```

### ⚠️ Breaking Change

1. **`Next` evaluation is deferred**  
   `c.Next()` now always runs **before** the skip check. This means the `Next` function receives the fully-processed context, including any data set by downstream middlewares or the handler. If your `Next` function previously expected an empty context, you may need to adjust it.

2. **Error handling still occurs when `Next` returns true**  
   In the old version, when `Next` returned `true` the middleware called `c.Next()` and immediately returned — the error handler was never triggered. Now errors from the handler chain are **always** handled (via `c.App().ErrorHandler`) before the skip decision. This is intentional: error handling is a framework concern and should not be bypassed by a logging skip. If you rely on the old behaviour (suppressing error handling when skipping logging), you can wrap the error handling block with the `Next` check, but this is not recommended.

3. **Return value when skipping**  
   When `cfg.Next` returns `true`, the middleware now returns `nil` instead of the result of a second `c.Next()` call. The handler chain has already completed; calling `c.Next()` again would re-execute it (which is both incorrect and dangerous). Returning `nil` signals that the middleware did not produce an error. If you need to preserve and propagate a downstream error, you can change the return value to `chainErr`, but note that the error has already been passed to the error handler, and returning it again may trigger duplicate error handling.

## How to Test
1. Register a middleware **after** the logger that sets a value via `c.Locals()`, e.g. `c.Locals("role", "admin")`.
2. In the logger's `Next` function, read `c.Locals("role")` and return `true` to skip logging.
3. Make a request that goes through both middlewares. Verify:
   - The request completes successfully.
   - No log output is produced.
   - The middleware does not panic or re-execute the handler.
4. Test with a handler that returns an error. Confirm:
   - The error handler is invoked.
   - Logging is skipped (if `Next` returns `true`), but the error is still properly handled.

## Additional Notes
- The logging logic (level selection, message selection, etc.) remains completely unchanged; only the control flow has been adjusted.
- This change is backwards incompatible in the ways described above. If you want to preserve the old behaviour, you can wrap your `Next` logic to check the context before it has been enriched, but updating your code to take advantage of the new order is the recommended path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Middleware now always executes the request chain and runs centralized error handling before evaluating the conditional "skip logging" rule, preventing downstream handlers from being bypassed and ensuring correct logging control flow.

* **Tests**
  * Updated tests to validate conditional logging: requests marked to skip produce no logs, while normal requests emit expected log entries including response body fields.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gofiber/contrib/pull/1939?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ReneWerner87:
We should not change the position for Next
In our normal logger middleware we introduced a DONE function 
https://docs.gofiber.io/middleware/logger

Maybe this is better and no breaking change

ReneWerner87:
close because of another PR https://github.com/gofiber/contrib/pull/1940

piyongcai-liucai:
> Reference in ne
**Reply to reviewer's suggestion about `Done`:**

Thanks for the suggestion! I understand the `Done` pattern from the built-in logger, but in this case it doesn't address the original performance concern.

The problem with using `Done` is that **it fires after the log message has already been constructed**. If I want to skip logging entirely based on some downstream context (e.g., user role), the log string will still be built and discarded, which is wasteful. The whole point of moving the `Next` check after `c.Next()` was to avoid that unnecessary work when logging is not needed.

Regarding middleware ordering: I'm aware that placing the logger *after* other middlewares would allow `Next` to read their context even in the current code. However, my concern is about **flexibility**. In many real-world applications, the logger is registered globally before route-specific middlewares that set context. Reordering is not always possible or clean, and the middleware should not impose such a strict ordering constraint.

I believe the minimal breaking change of moving `Next` after `c.Next()` is justified:
- It allows `Next` to access the fully resolved context regardless of ordering.
- It still preserves error handling (which I've kept in the updated code).
- The `Next` function is explicitly designed to skip logging, so deferring it to a point where all context is available is logical.

If a `Done` callback is also desired for post-logging actions, it could be added as a separate non-breaking enhancement, but it doesn't solve the skip-logging use case.

Would you be open to reconsidering moving the `Next` check, or is there another way to achieve zero-cost skipping while keeping the original middleware order?

piyongcai-liucai:
> close because of another PR #1940

Thank you for the feedback. I’d like to clarify the situation:

- **#1940** was created by mistake — I intended to target `v2` but accidentally selected `main`. I’ve closed it now.
- **#1939** is the correct PR I want to keep. It is intentionally targeting `main` and contains the intended changes.

Could you please reopen **#1939**? I’ll make sure it’s ready for review. Sorry for the confusion and thank you for your time.

ReneWerner87:
Okay, now I understand the problem.
You want to respond to the work that the middleware has done in the meantime.

Hmm, should we move that into our logger middleware as well?

https://github.com/gofiber/fiber/blob/main/middleware/logger/logger.go#L82-L86

Or should we handle it differently with another callback for a different point in time?
WDYT @gofiber/maintainers

piyongcai-liucai:
> Okay, now I understand the problem. You want to respond to the work that the middleware has done in the meantime.
> 
> Hmm, should we move that into our logger middleware as well?
> 
> https://github.com/gofiber/fiber/blob/main/middleware/logger/logger.go#L82-L86
> 
> Or should we handle it differently with another callback for a different point in time? WDYT @gofiber/maintainers

Thank you for looking into this and understanding the problem.

I'd like to clarify my thoughts on the approaches discussed:

1. Regarding the **`Done` callback** — while it works well in the built-in logger, it fires **after** the log message has already been assembled. If the goal is to skip logging entirely, the performance cost has already been incurred. So I don't think it's the ideal solution for this specific use case.

2. I believe there are two better approaches, and I'm happy to go with whichever the @gofiber/maintainers prefer:
   - **a)** Port my `Next`-after-`c.Next()` pattern into the built-in logger middleware, so the skip decision can be made with full context **before** any log assembly happens.
   - **b)** Design a new callback that fires at a different point in time (after `c.Next()` but before logging), which is also a great solution.

Both options avoid the unnecessary performance overhead, and I'll implement whichever direction the team decides. Thanks again for your time!

gaby:
@ReneWerner87 This should be a config option. Basically a `Lazy Validation`.

If true, do Next check at the end, else keep current behavior.

gaby:
@piyongcai-liucai let us discuss the feature request first.

ReneWerner87:
@coderabbitai You're familiar with our middleware configurations from all the repositories, i.e., the core and contrib middleware packages.
What do you think is the best way to enable an option to skip at this specific point in time?
Please list the options so we can vote on them.

Additional question: Could this also be solved using the skip middleware?
This middleware issue also applies to our logger middleware, because the exact same problem exists there, and this one has a similar function but with different adapters.

piyongcai-liucai:
I'd like to cast my vote for **Option A** — the `Skip func(c fiber.Ctx) bool` callback.

Reasons:
- **Zero-cost skip**: It runs after `c.Next()` but before log assembly, so no performance is wasted when skipping.
- **Non-breaking**: It's a new optional field; existing `Next` behavior remains untouched.
- **Consistent with core logger**: The built-in Fiber logger already ships with this pattern, making it familiar to users and maintainers.

It solves my original use case perfectly. Thanks for considering!

ReneWerner87:
@piyongcai-liucai i check together with some of the maintainers and we are okay with option A -> can you change your PR for this

piyongcai-liucai:
> @piyongcai-liucai i check together with some of the maintainers and we are okay with option A -> can you change your PR for this

Done — I've updated the PR to implement **Option A**.

Changes:
- Added `Skip func(c fiber.Ctx) bool` field to `Config`.
- The `Skip` callback runs after `c.Next()` but before log assembly, allowing a zero-cost skip with full request context.
- The existing `Next` behavior remains untouched for full backward compatibility.

This mirrors the `Skip` pattern used in the core logger middleware. Please take another look when you have a moment. Thanks for the guidance!

ReneWerner87 (changes requested):
@piyongcai-liucai Please add information to the documentation (md) files.
A feature without documentation and examples always leads to a lot of questions or to the feature not being used.

ReneWerner87:
@piyongcai-liucai ping

piyongcai-liucai:
Thanks for the feedback. I've added the `Skip` callback to the Config table in the README, along with a usage example and a comparison between `Skip` and `Next`. Let me know if any further adjustments are needed.

ReneWerner87:
@piyongcai-liucai 
> along with a usage example and a comparison between Skip and Next. Let me know if any further adjustments are needed.

i dont see the example in the markdown ?

ReneWerner87 (approved):

PR #1989, opened 2026-06-08
feat: Add static window interval support

igneel64:
This pull request adds support for basic interval window in the circuit breaker middleware, allowing users to choose an interval in which if in closed state, the failure counts would be reset.

My thinking is that if this is an acceptable way to go, I can port this to v3 and later add rolling window support as other circuit breaker solutions have.

- [x] Added a few tests
- [x] Updated documentation with example usage
- [x] No new dependency introduced


ReneWerner87:
sorry no new features for v2 -> there is a feature stop

igneel64:
@ReneWerner87 Thank you for the nudge, didn't know, opened up for v3 :)

PR #2038, opened 2026-06-26
🔥 feat: Add Uptime Middleware

gofurry:
This PR adds a new `uptime` middleware for Fiber v3.

The middleware records in-process heartbeat history, stores uptime state through `github.com/gofiber/storage/redis/v3`, and serves a built-in status dashboard with a JSON status API.

QQ20260626-182538

## Summary by CodeRabbit

* **New Features**
  * Added Fiber v3 uptime middleware that records periodic heartbeats, performs daily rollups and retention cleanup, and serves a `/api/status` JSON endpoint plus an auto-refreshing interactive dashboard with live/stale/error indicators.
* **Documentation**
  * Added comprehensive uptime package documentation and updated the middleware listings in the main and v3 READMEs.
* **Tests**
  * Added extensive automated tests covering configuration validation, handler behavior (GET/HEAD, routing, error surfacing), status/slot calculations, and Redis storage lifecycle.


v3/uptime/config.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	if c.UI.GreenThreshold == 0 {
+		c.UI.GreenThreshold = defaultGreenThreshold
+	}
+	if c.UI.YellowThreshold == 0 {
+		c.UI.YellowThreshold = defaultYellowThreshold
+	}
+	if c.UI.GreenThreshold < c.UI.YellowThreshold {
+		return Config{}, errors.New("uptime: green threshold must be greater than or equal to yellow threshold")
+	}
```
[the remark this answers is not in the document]
gofurry: Fixed. Added [0,1] bounds validation for both UI thresholds during config normalization, with test coverage.

v3/uptime/internal/storage/sqlite.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+	err := s.db.Close()
+	s.db = nil
+	return err
+}
+
+func ensureSQLiteDir(path string) error {
+	if path == ":memory:" || strings.HasPrefix(path, "file:") {
+		return nil
+	}
```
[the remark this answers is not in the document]
gofurry: Fixed. SQLite file: URIs now strip the scheme/query and create parent directories for disk-backed paths. Added a regression test.

v3/uptime/README.md
```diff
@@ -0,0 +1,169 @@
...
+![Test](https://github.com/gofiber/contrib/workflows/Test%20Uptime/badge.svg)
+
+Uptime middleware for [Fiber](https://github.com/gofiber/fiber) that records
+in-process heartbeat history and serves a lightweight status page.
+
+**Compatible with Fiber v3.**
+
+## Preview
+
+![Uptime dashboard preview](https://raw.githubusercontent.com/gofurry/images/refs/heads/main/github/uptime/preview.png)
```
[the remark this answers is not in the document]
gofurry: Not changed. I put the preview picture there.

v3/uptime/status.go
```diff
@@ -0,0 +1,433 @@
...
+	for _, row := range todayRows {
+		todayByService[row.ServiceID] = row
+	}
+
+	resp := StatusResponse{
+		GeneratedAt:           now.UTC(),
+		SampleIntervalSeconds: int64(u.config.SampleInterval / time.Second),
+		Days:                  u.config.DaysToShow,
+		Storage:               u.storageStatus(),
+		Services:              make([]ServiceStatus, 0, len(services)),
```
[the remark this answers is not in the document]
gofurry: Not applied as suggested. LastError also tracks heartbeat/maintenance runtime errors, so a successful read-only status snapshot should not clear recorder-side failures.

v3/uptime/ui.go
```diff
@@ -0,0 +1,922 @@
...
+  transition: background-color 120ms ease, filter 120ms ease;
+}
+.bar:hover,
+.bar:focus-visible,
+.bar.is-active {
+  outline: none;
+  box-shadow: none;
+  transform: none;
+  filter: saturate(1.04) brightness(.9);
+}
```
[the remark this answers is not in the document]
gofurry: Fixed. :focus-visible now has a visible outline separate from hover/active styling.

v3/uptime/ui.go
```diff
@@ -0,0 +1,922 @@
...
+    <div class="header-divider" aria-hidden="true"></div>
+    <aside class="description-card" aria-label="Page description">
+      <p>{{ .Description }}</p>
+    </aside>
+
+    <section class="services" id="services" aria-label="Services"></section>
+    <footer>{{ .Footer }}</footer>
+  </main>
+
+  <button id="page-scroll-dock" class="page-scroll-dock" type="button" title="Scroll up" aria-label="Scroll progress 0%" aria-valuemin="0" aria-valuemax="100" aria-valuenow="0">
```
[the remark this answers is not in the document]
gofurry: Fixed. The scroll dock aria-label now includes both the scroll-up action and current progress.

v3/uptime/ui.go
```diff
@@ -0,0 +1,922 @@
...
+};
+window.uptimeConfig = { apiPath, refreshMS };
+window.uptimeInitialStatus = initialStatus;
+
+const $ = (id) => document.getElementById(id);
+let currentStatus = "live";
+let lastStatus = initialStatus;
+let lastSuccessAt = initialStatus ? Date.now() : 0;
+let activeBar = null;
+
```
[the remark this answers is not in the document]
gofurry: Fixed. Dashboard refresh now has an in-flight guard and an AbortController timeout.

v3/uptime/uptime_test.go
```diff
@@ -0,0 +1,667 @@
...
+	requireEqual(t, fiber.StatusOK, resp.StatusCode)
+	requireEqual(t, fiber.MIMETextHTMLCharsetUTF8, resp.Header.Get(fiber.HeaderContentType))
+	requireContains(t, bodyText, "<title>Fiber Uptime</title>")
+	requireContains(t, bodyText, `const apiPath = "/uptime/api/status";`)
+	requireContains(t, bodyText, `id="summary-services"`)
+	requireContains(t, bodyText, `id="summary-up"`)
+	requireContains(t, bodyText, `id="summary-down"`)
+	requireContains(t, bodyText, `const refreshMS =`)
+	requireContains(t, bodyText, `10000`)
+	requireContains(t, bodyText, `date.getFullYear() + "-" + pad(date.getDate()) + "-" + pad(date.getMonth() + 1)`)
```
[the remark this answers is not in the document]
gofurry: Fixed. The dashboard formatter now uses the standard YYYY-MM-DD HH:mm:ss format, with test coverage updated.

v3/uptime/uptime.go
```diff
@@ -0,0 +1,377 @@
...
+
+	store := storage.NewSQLiteStore(cfg.SQLite)
+	return newWithStore(cfg, store, time.Now())
+}
+
+func newWithStore(cfg Config, store storage.Store, now time.Time) (*Uptime, error) {
+	ctx := context.Background()
+	if err := store.Init(ctx); err != nil {
+		return nil, fmt.Errorf("uptime: init store: %w", err)
+	}
```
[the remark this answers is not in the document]
gofurry: Fixed. The store is now closed when Init fails during middleware construction.

v3/uptime/uptime.go
```diff
@@ -0,0 +1,377 @@
...
+}
+
+func (u *Uptime) serveStatusJSON(c fiber.Ctx) error {
+	c.Set(headerCacheControl, "no-store")
+	c.Set(headerXContentTypeOptions, "nosniff")
+	c.Set(fiber.HeaderContentType, fiber.MIMEApplicationJSONCharsetUTF8)
+	if c.Method() == fiber.MethodHead {
+		c.Status(fiber.StatusOK)
+		return nil
```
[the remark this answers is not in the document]
gofurry: Fixed. HEAD requests now exercise the same snapshot/render error paths as GET, while still returning an empty body on success.

v3/uptime/ui.go
```diff
@@ -0,0 +1,938 @@
...
+}
+
+func maxInt64(a, b int64) int64 {
+	if a > b {
+		return a
+	}
+	return b
+}
+
+const dashboardHTML = `<!doctype html>
```
ReneWerner87: @gofurry can you outsource this in a real file

ReneWerner87:
@gofurry pls check https://github.com/gofiber/contrib/pull/2038#discussion_r3481415529 
and also add dark mode based on the system settings

v3/uptime/ui.go
gaby: I think it has to be this way, i ran into the same thing with openapi middleware

gaby: It can only be in a file if static, if you are replacing variables is has to be a template.

gaby:
@gofurry I would suggest using https://github.com/gofiber/storage/tree/main/sqlite3 to create the sqlite db/connection and the using the `Conn()` to run custom queries against it. Instead of creating a full custom sqlite storage driver

v3/uptime/ui.go
gofurry: Moved the dashboard body into an embedded dashboard.gohtml file while keeping the same Go html/template rendering path.

gofurry:
> @gofurry pls check [#2038 (comment)](https://github.com/gofiber/contrib/pull/2038#discussion_r3481415529) and also add dark mode based on the system settings

Added system-based dark mode using prefers-color-scheme, without adding a theme switcher. The dashboard now keeps the light UI as the default and automatically adapts to the user's system preference.

QQ20260627-140647

gofurry:
> @gofurry I would suggest using https://github.com/gofiber/storage/tree/main/sqlite3 to create the sqlite db/connection and the using the `Conn()` to run custom queries against it. Instead of creating a full custom sqlite storage driver

Thanks for the suggestion. I checked `github.com/gofiber/storage/sqlite3/v2` and it currently uses `github.com/mattn/go-sqlite3` under the hood.

Before I switch the uptime storage to use it via `Conn()`, I want to confirm whether introducing a CGO requirement is acceptable for this middleware. Moving to `gofiber/storage/sqlite3/v2` would make consumers of the default SQLite-backed uptime middleware require CGO for builds.

v3/uptime/dashboard.gohtml
```diff
@@ -0,0 +1,927 @@
...
+      <p>{{ .Description }}</p>
+    </aside>
+
+    <section class="services" id="services" aria-label="Services"></section>
+    <footer>{{ .Footer }}</footer>
+  </main>
+
+  <button id="page-scroll-dock" class="page-scroll-dock" type="button" title="Scroll up" aria-label="Scroll up; scroll progress 0%" aria-valuemin="0" aria-valuemax="100" aria-valuenow="0">
+    <span id="page-scroll-dock-value" class="page-scroll-dock__core">0%</span>
+  </button>
```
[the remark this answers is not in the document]
gofurry: Fixed.

v3/uptime/dashboard.gohtml
```diff
@@ -0,0 +1,927 @@
...
+  dock.classList.toggle("page-scroll-dock--visible", shouldShow);
+  value.textContent = rounded + "%";
+}
+
+function scrollUpQuarter() {
+  const doc = document.documentElement;
+  const maxScroll = Math.max(0, doc.scrollHeight - window.innerHeight);
+  const top = window.scrollY || doc.scrollTop || document.body.scrollTop || 0;
+  window.scrollTo({ top: Math.max(0, top - maxScroll * 0.25), behavior: "smooth" });
+}
```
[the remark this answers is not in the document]
gofurry: Fixed.

gaby:
> > @gofurry I would suggest using https://github.com/gofiber/storage/tree/main/sqlite3 to create the sqlite db/connection and the using the `Conn()` to run custom queries against it. Instead of creating a full custom sqlite storage driver
> 
> Thanks for the suggestion. I checked `github.com/gofiber/storage/sqlite3/v2` and it currently uses `github.com/mattn/go-sqlite3` under the hood.
> 
> Before I switch the uptime storage to use it via `Conn()`, I want to confirm whether introducing a CGO requirement is acceptable for this middleware. Moving to `gofiber/storage/sqlite3/v2` would make consumers of the default SQLite-backed uptime middleware require CGO for builds.

Let's keep it as is then.

v3/uptime/README.md
```diff
@@ -0,0 +1,169 @@
...
+import (
+	"github.com/gofiber/contrib/v3/uptime"
+	"github.com/gofiber/fiber/v3"
+	"github.com/gofiber/fiber/v3/log"
+)
+
+func main() {
+	app := fiber.New()
+
+	up, err := uptime.New(uptime.Config{
```
gaby: This setup is very anti-fiber pattern. It looks awful. The middlewares are aupposed to be handlers.

Closing the DB can be done using a signal handler or hooks

gaby:
Can't we use Redis or something else from gofiber/storage? I'm not a huge fan of the massive database implementation here.

Ideally Redis.

v3/uptime/README.md
gaby: This is a perfect reason not to have a custom DB setup.

Normally to create a DB conn outside of the fiber app and pass it to the app, so you can close in on a hook.

gaby:
- Remove the snapshot logic or simplify it, it is making the code very convoluted, we have a cache middleware. Most database have built-in cache, we don't need to reimplement that.

- How do we configure what services/endpoints to monitor?

gofurry:
> * Remove the snapshot logic or simplify it, it is making the code very convoluted, we have a cache middleware. Most database have built-in cache, we don't need to reimplement that.
> * How do we configure what services/endpoints to monitor?

Addressed. `uptime.New(config...)` now returns a `fiber.Handler`, so it can be registered like the other contrib middlewares. The custom SQLite storage has been removed and the middleware now uses `github.com/gofiber/storage/redis/v3`. The snapshot cache layer was simplified, and configurable endpoint probes were added. Each endpoint is tracked as a service target and probed in the background with its own method, URL, timeout, expected status codes, and optional headers.

v3/uptime/README.md
```diff
@@ -0,0 +1,225 @@
...
+`StorageKeyPrefix` when multiple environments share the same Redis server.
+
+## Snapshots and custom UI
+
+The dashboard and JSON API build a fresh `Snapshot` from the backing store on
+each request. Use Fiber's cache middleware around the uptime route if you want
+HTTP-level caching.
+
+```go
+up, err := uptime.NewRuntime(uptime.Config{
```
gaby: We should only have a handler that takes an UptimeConfig. The config takes a Fiber Redis Storage client.

The storage drivers by default will panic if redis is Down. That way you don't have to be creating clients and returning errors or doing New().

gaby: @gofurry everything else looks fine, but the usage of the middleware should be as simple as possible. Right now it is a multi step process.

gofurry: The `Config` in `uptime.New(config...)` accepts a Fiber Redis storage client via `Store`, and the middleware no longer creates/configures Redis internally

v3/uptime/README.md
gofurry: Addressed.

v3/uptime/internal/storage/redis.go
```diff
@@ -0,0 +1,617 @@
...
+			UpSlots:   int(upSlots),
+		})
+	}
+	return statuses, nil
+}
+
+func (s *RedisStore) Close() error {
+	s.storage = nil
+	s.client = nil
+	return nil
```
gaby, line 502: **Panic + data race: `Close()` nils the client while request handlers may still be using it.**

`Close()` sets `s.client = nil` with no synchronization, but Fiber v3 runs `OnPreShutdown` hooks *before* the server stops accepting/draining requests (`ShutdownWithContext` executes the hooks at app.go:1267 before `server.ShutdownWithContext` at :1270). `runtime.close()` only waits for the recorder goroutines (`<-u.done`), not for in-flight HTTP handlers, so during `app.Shutdown()` an in-flight `GET /uptime/api/status` can reach `ListServices` → `s.client.SMembers(...)` on a nil interface and panic the process; `-race` also flags the unsynchronized write/read of `s.client`.

Since the config docs say the caller owns the Redis storage lifecycle, the simplest fix is for `Close()` to not nil the fields at all (or guard `client` with a mutex / atomic and have callers return an error when it's gone).

---

v3/uptime/internal/storage/redis.go
```diff
@@ -0,0 +1,617 @@
...
+	if err != nil {
+		return err
+	}
+	for _, service := range services {
+		days, err := s.daysBefore(ctx, s.sampleDaysKey(service.ID), options.BeforeDay)
+		if err != nil {
+			return err
+		}
+		for _, day := range days {
+			upSlots64, err := s.client.SCard(ctx, s.sampleKey(service.ID, day)).Result()
```
gaby, line 278: **Multi-instance race: rollup vs. cleanup can permanently zero a finalized day.**

Maintenance is guarded only by the process-local `u.maintenanceMu` — there is no cross-process coordination (no `SET NX` lease/Lua lock), and multi-instance on shared Redis is this module's documented topology. Consider two instances after a ≥2-day gap (both run forced maintenance on startup, uptime.go:133):

1. Instance A's `RollupDaily` fetches its days list including day `D` (≤ today-2, still in `sample-days`).
2. Instance B's `Cleanup` runs `DEL samples(D)` + `ZREM`.
3. A's `SCard` on the deleted key returns `0`, and `writeDaily` unconditionally overwrites the previously-correct daily hash with `up_slots=0, uptime_rate=0, finalized=1`. A's own cleanup then ZREMs `D`, so it is never recomputed — the day renders as a permanent 0% red bar that was actually fine.

Related milder effects of the same design: every instance re-rolls yesterday every minute all day, and instances with different `Timezone`/`SampleInterval` configs last-writer-win on `expected_slots` of "finalized" rows. A Redis-side lock (`SET key val NX PX ...`) around the rollup+cleanup pass, or skipping `writeDaily` when the day is no longer in `sample-days`, would close the window.

---

v3/uptime/uptime.go
```diff
@@ -0,0 +1,549 @@
...
+			return fiber.NewError(fiber.StatusInternalServerError, "uptime unavailable")
+		}
+		if u.config.Next != nil && u.config.Next(c) {
+			return c.Next()
+		}
+
+		path := c.Path()
+		uiPath := u.config.UI.Path
+		route := ""
+		switch path {
```
gaby: **Dashboard silently unreachable when mounted under a group or prefix.**

`c.Path()` in Fiber v3 returns the full request path with no prefix stripping, so with `app.Group("/admin").Use(uptime.New(...))` or `app.Use("/admin", uptime.New(...))` the handler sees `/admin/uptime`, which matches neither `uiPath` nor the `uiPath+"/"` prefix branch — every request falls through to `c.Next()` and the dashboard/API 404 with no hint why (verified empirically against fiber v3.3.0). Conversely, at root the `strings.HasPrefix` branch 404-hijacks the entire `UI.Path+"/"` namespace, shadowing any user route registered under it, and ignores the router's `CaseSensitive`/`StrictRouting` settings.

Since `Config.App` is already available, registering real routes (`GET`/`HEAD` on `uiPath` and `uiPath+"/api/status"`) would give correct matching, 405-with-`Allow`, and group mounting for free — or at minimum compare against a path relative to the mount point and document that the middleware must be mounted at root.

---

v3/uptime/uptime.go
```diff
@@ -0,0 +1,549 @@
...
+	headerCacheControl        = "Cache-Control"
+	headerXContentTypeOptions = "X-Content-Type-Options"
+	headerAllow               = "Allow"
+)
+
+// New creates a Fiber handler that records uptime and serves the dashboard/API.
+func New(config ...Config) fiber.Handler {
+	u, err := newRuntime(config...)
+	if err != nil {
+		panic(fmt.Errorf("fiber: uptime middleware error -> %w", err))
```
gaby, line 79: **`New()` panics on any transient Redis failure at startup, although the middleware already has a degraded mode.**

`newWithStore` performs blocking network I/O before returning: `Ping`, `UpsertService`/`UpsertInstance` per target, a full forced maintenance pass, and the initial heartbeat — any error bubbles up to this `panic`. A pod that starts a few seconds before Redis is reachable (or during a Redis failover) crash-loops the whole application, and a rolling restart of many replicas turns a transient blip into an outage.

The same errors at steady state are handled gracefully (`setLastError` → `storage.status: "degraded"` in the API). Panicking only on genuine config errors and letting the store initialize lazily/retried from the background loop would make startup no more fragile than runtime.

---

v3/uptime/uptime.go
```diff
@@ -0,0 +1,549 @@
...
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+	}
+
+	if target.probe != nil && !u.probeEndpoint(ctx, target.probe) {
+		if err := u.runMaintenance(ctx, now, false); err != nil {
+			return err
+		}
+		u.clearLastError()
```
gaby: **Failed-probe path calls `clearLastError()` without any storage operation succeeding, masking real Redis outages.**

When a probe fails, `runMaintenance` usually early-returns `nil` without touching Redis (same day, < 1 minute since last run), and then `clearLastError()` wipes whatever error another target just recorded. Concrete scenario: Redis becomes unreachable → the service target's `WriteHeartbeat` fails and `setLastError` makes the API report `storage: "degraded"`; within the next probe interval a down endpoint takes this branch and resets the error → the dashboard shows storage `"ok"` while no heartbeats are being persisted at all. For a monitoring middleware this hides exactly the outage it exists to surface.

`clearLastError()` should only run after an operation that actually talked to the store successfully (e.g. drop it from this branch; heartbeat writes and maintenance already clear it on success).

---

v3/uptime/status.go
```diff
@@ -0,0 +1,398 @@
...
+		return DayStatus{
+			Day:     day,
+			HasData: false,
+			Status:  "gray",
+		}
+	}
+
+	if day == today {
+		row := todayRows[serviceID]
+		expected := expectedSlotsSoFarSince(now, createdAt, interval, u.config.Timezone)
```
gaby, line 183: **Healthy services render red/yellow every morning: today's expected slots include the in-progress slot, but the heartbeat ticker isn't slot-aligned.**

`expectedSlotsSoFarSince` uses `includeEnd=true`, so `lastSlot = slotOf(now)` — the slot that just started counts as expected. Heartbeats come from a `time.NewTicker` phased to process start (uptime.go:364), not to the midnight-anchored slot boundaries, so the newest slot frequently has no heartbeat yet. With the defaults (3s interval, green ≥ 0.999, yellow ≥ 0.99):

- at 00:00:01, expected = 1, up = 0 → today's bar is **red at 0%**;
- at 00:00:30, ~10/11 ≈ 90.9% → still red;
- one lagging slot keeps the rate below green until ~1000 expected slots, i.e. today reads yellow/red for roughly the first 50 minutes of every day on a service with 100% uptime.

Nothing mitigates it (`hasData` is hardcoded `true` for today, no grace slot or young-day handling). Excluding the current in-progress slot from the expected count (`includeEnd=false`, or `lastSlot = slotOf(now) - 1` when now isn't past the slot's end) would fix the daily false alarm.

---

v3/uptime/uptime.go
```diff
@@ -0,0 +1,549 @@
...
+
+	runCtx, cancel := context.WithCancel(context.Background())
+	u := &runtime{
+		config:     cfg,
+		store:      store,
+		targets:    targets,
+		ctx:        runCtx,
+		cancel:     cancel,
+		done:       make(chan struct{}),
+		httpClient: &http.Client{},
```
gaby: **Probe client follows redirects, so `ExpectedStatusCodes` containing a 3xx can never match.**

The default `&http.Client{}` transparently follows up to 10 redirects. An endpoint configured as `ExpectedStatusCodes: []int{301}` gets the *final* response (e.g. 200) from the redirect target, `statusAllowed(200)` is false, and the endpoint is recorded down on every probe — 0% uptime for a healthy redirect. It also means the probe silently measures the redirect target rather than the configured URL, and the default "any 2xx or 3xx is up" rule (config.go:96) can likewise never observe an actual 3xx.

Setting `CheckRedirect: func(*http.Request, []*http.Request) error { return http.ErrUseLastResponse }` makes the probe evaluate the configured URL's own status code.

---

v3/uptime/uptime.go
```diff
@@ -0,0 +1,549 @@
...
+		timeout:             endpoint.Timeout,
+		expectedStatusCodes: expectedStatusCodes,
+	}
+}
+
+func registerShutdownHook(app *fiber.App, u *runtime) {
+	if app == nil || u == nil {
+		return
+	}
+	app.Hooks().OnPreShutdown(u.close)
```
gaby, line 264: **Without `Config.App` there is no way to stop the runtime — goroutines leak and the service keeps reporting "up" after shutdown.**

`App` is documented as optional, but the shutdown hook is the *only* path to `u.close()`; `New()` returns just a `fiber.Handler`. When `App` is unset and the caller shuts their server down (or in test suites that construct the middleware repeatedly), the per-target `recordLoop` goroutines, tickers, and the HTTP client keep running for the life of the process and keep writing heartbeats to Redis — so the dashboard shows a service that no longer serves traffic as up with 100% uptime indefinitely.

Consider making `App` required, or providing an explicit stop mechanism (e.g. return a closer, or lazily start the recorder on first request), and documenting the leak if `App` stays optional.

---

v3/uptime/internal/storage/redis.go
```diff
@@ -0,0 +1,617 @@
...
+		"name", service.Name,
+		"description", service.Description,
+		"sample_interval_nanos", int64(service.SampleInterval),
+	).Err(); err != nil {
+		return err
+	}
+	return s.setMaxNano(ctx, serviceKey, "last_seen_at", unixNano(service.LastSeenAt))
+}
+
+func (s *RedisStore) UpsertInstance(ctx context.Context, instance Instance) error {
```
gaby, line 166: **Instance hashes leak in Redis forever — written on every boot, never read, never cleaned up.**

`instanceIDForNode` generates a fresh pseudo-random ID on each process start, and `UpsertInstance` writes a `<prefix>:instance:<id>` hash (plus one per endpoint target) with no TTL. `Cleanup` only touches daily/sample keys, and no query ever reads instance hashes back. A service that restarts regularly (deploys, autoscaling) accumulates one orphaned hash per target per boot, unbounded, and the data is never consumed by anything.

Either drop instance persistence entirely (nothing uses it), or give instance keys a TTL / index them so cleanup can reap stale ones. Related: `buildRecordTargets` derives endpoint instance IDs as `cfg.InstanceID + index`, so two processes configured with adjacent explicit `InstanceID`s (100 and 101) silently write the same instance hash.

---

.github/workflows/test-uptime.yml
```diff
@@ -0,0 +1,52 @@
+name: "Test Uptime"
+
+on:
+  push:
+    branches:
+      - master
+      - main
+    paths:
+      - 'v3/uptime/**/*.go'
```
gaby, line 9: **Paths filter misses `dashboard.gohtml`, which is `go:embed`-ed and load-bearing for every test.**

`ui.go` embeds the template and parses it at package init via `template.Must`, so a PR that edits only `v3/uptime/dashboard.gohtml` (e.g. introduces a typo'd `{{.Field}}` action) matches none of these path filters — "Test Uptime" never runs and the PR merges green, then `template.Must` panics at init (or `Execute` fails at request time) for every consumer. Add the template to both `push` and `pull_request` filters:

```suggestion
      - 'v3/uptime/**/*.go'
      - 'v3/uptime/dashboard.gohtml'
```

(and the same entry under the `pull_request` paths below).

---

gaby (commented):
Automated high-effort code review of the new `uptime` middleware. Candidates were gathered across 8 independent review angles and each surviving finding was independently verified against the code (and fiber v3.3.0 / fasthttp sources where framework behavior mattered); several were reproduced empirically with throwaway tests. 10 findings, most severe first:

1. **Shutdown panic/race** — `RedisStore.Close()` nils the client while in-flight handlers can still call it; Fiber runs `OnPreShutdown` before draining requests (`redis.go:369`).
2. **Multi-instance data loss** — concurrent rollup/cleanup across instances can permanently zero a correctly-finalized day (`redis.go:169`).
3. **Group mounting silently breaks** — `c.Path()` includes the group prefix, so `app.Group("/admin").Use(uptime.New(...))` makes the dashboard 404 (`uptime.go:274`).
4. **Startup fragility** — `New()` panics on transient Redis errors (ping/upserts/initial heartbeat/maintenance) even though a degraded mode exists (`uptime.go:74`).
5. **Outage masking** — the failed-probe path clears `lastErr` without any successful storage operation, flipping `storage: "degraded"` back to `"ok"` during a Redis outage (`uptime.go:388`).
6. **False red bars** — today's expected slots include the in-progress slot while the heartbeat ticker isn't slot-aligned; healthy services render red/yellow for ~50 min after every midnight with defaults (`status.go:169`).
7. **3xx probes impossible** — the probe `http.Client` follows redirects, so `ExpectedStatusCodes: []int{301}` can never match (`uptime.go:130`).
8. **Unstoppable runtime** — without optional `Config.App` there is no close path; goroutines leak and a stopped app keeps reporting "up" (`uptime.go:231`).
9. **Redis key leak** — instance hashes are written every boot with no TTL, never read, never cleaned (`redis.go:116`).
10. **CI gap** — the workflow paths filter misses the `go:embed`-ed `dashboard.gohtml` (`test-uptime.yml:9`).

Notable items that didn't make the top-10 cut: HEAD responses lose the entity `Content-Length` (fasthttp would suppress the body automatically if the handlers just wrote it); `QueryDaily`/`QueryTodaySamples` each re-call `ListServices` so one snapshot fetches the service list three times; the persisted `uptime_rate` field is never read back (status.go recomputes it) and `rate()` duplicates `uptimeRate()`; explicit `GreenThreshold`/`YellowThreshold` of `0` is silently replaced by defaults despite docs saying the range is `[0, 1]`; and the README example registers `store.Close` before the runtime's shutdown hook, so final heartbeats hit a closed client.

---

v3/uptime/internal/storage/redis.go
gofurry: The Redis store is no longer closed or nilled by the middleware.

v3/uptime/internal/storage/redis.go
gofurry: Daily rollups now preserve already-finalized rows, so a later concurrent cleanup/rollup cannot overwrite a correct finalized day.

v3/uptime/uptime.go
gofurry: The handler now supports group/prefix mounted paths and unknown uptime subpaths pass through to the next handler.

v3/uptime/uptime.go
gofurry: Startup Redis I/O failures now start the middleware in degraded mode instead of panicking; config errors remain fail-fast.

v3/uptime/uptime.go
gofurry: Failed endpoint probes no longer clear existing storage errors.

v3/uptime/status.go
gofurry: Today’s expected slots now exclude the in-progress slot.

v3/uptime/uptime.go
gofurry: Probe redirects are no longer followed, so configured 3xx status codes can be evaluated.

v3/uptime/uptime.go
gofurry: Config.App is now required so the runtime always has a Fiber shutdown hook.

v3/uptime/internal/storage/redis.go
gofurry: Redis instance metadata now has expiration/cleanup.

.github/workflows/test-uptime.yml
gofurry: The uptime workflow now includes dashboard.gohtml in its path filters.

v3/uptime/ui.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return "", err
+	}
+	page := dashboardPage{
+		Title:       config.UI.Title,
+		Description: config.UI.Description,
+		Footer:      config.UI.Footer,
+		APIPathJSON: template.JS(apiPathJSON),
+		RefreshMS:   max(int64(config.SampleInterval/time.Millisecond), 10000),
+		StatusJSON:  template.JS(statusJSON),
```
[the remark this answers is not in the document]
gofurry: Dashboard JSON is converted to string before casting to template.JS.

v3/uptime/config.go
```diff
@@ -0,0 +1,325 @@
...
+	Path string
+	// Description is shown below the dashboard header.
+	Description string
+	// Footer is shown at the bottom of the dashboard.
+	Footer string
+
+	// GreenThreshold is the minimum uptime ratio for a green day, in the range [0, 1].
+	GreenThreshold float64
+	// YellowThreshold is the minimum uptime ratio for a yellow day, in the range [0, 1].
+	YellowThreshold float64
```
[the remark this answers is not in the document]
gofurry: Threshold docs now clarify that configurable values are in (0, 1], and zero uses defaults.

v3/uptime/config.go
```diff
@@ -0,0 +1,325 @@
...
+	// InstanceID overrides the generated process instance ID when non-zero.
+	InstanceID int64
+	// IDGenerator generates instance IDs when InstanceID is zero.
+	IDGenerator IDGenerator
+
+	// Store is the Fiber Redis storage instance used for uptime state.
+	// The caller owns the storage lifecycle and should close it when the app stops.
+	Store *fiberredis.Storage
+	// StorageKeyPrefix namespaces all uptime keys inside the selected Redis database.
+	StorageKeyPrefix string
```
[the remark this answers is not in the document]
gofurry: Addressed.

v3/uptime/uptime.go
```diff
@@ -0,0 +1,646 @@
...
+	for _, target := range targets {
+		if err := store.UpsertService(ctx, target.service); err != nil {
+			_ = store.Close()
+			return nil, fmt.Errorf("uptime: upsert service: %w", err)
+		}
+		if err := store.UpsertInstance(ctx, target.instance); err != nil {
+			_ = store.Close()
+			return nil, fmt.Errorf("uptime: upsert instance: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
gaby: @gofurry we should be close to merging after this fix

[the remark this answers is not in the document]
gofurry: Target registration now goes through the existing startup path. This avoids using the Redis client before `Init()` while preserving degraded startup behavior when Redis is temporarily unavailable.

gaby:
@ReneWerner87 Have you run this locally to see how it looks/behaves?

ReneWerner87:
not really, should i do this ?

ReneWerner87:
image
image

ReneWerner87:
I ran this locally against a real Redis (screenshots above): one process heartbeat plus three probed endpoints, one healthy, one flaky, one always failing. The dashboard and `/api/status` behave as documented.

Two things I hit while doing that, both now fixed in `baa51656`:

**Services could never be removed.** `buildStatus` lists everything under the key prefix, not the configured services, and the Redis client interface had no `SRem`, so nothing could drop one. I renamed an endpoint ID and restarted: the old ID kept its dashboard row, reporting down, with the rollup writing it zero-days every minute. Only a manual Redis edit could clear it. Added `uptime.RemoveService`.

**Only instance keys had a TTL.** Everything else relied entirely on the maintenance loop still running, so a retired service or a changed key prefix leaked keys permanently. Added a `RetentionDays + 1` backstop, re-armed on every write, so cleanup still wins while a process is alive. That surfaced a follow-on case: a day whose samples expired while still in the index would roll up as a finalized zero, which the guard script then refuses to correct. Empty set now means expired, so the stale index entry is dropped instead.

`cba20d88` is docs only.

@gofurry sorry for pushing straight to your branch, `RemoveService` is a new exported function so please say if you'd rather shape that API differently.

Open points I did **not** change, all judgment calls rather than bugs:

- Redis is effectively mandatory. The middleware never uses the `fiber.Storage` interface, only `Conn()` for the raw go-redis client, because the model needs sets, sorted-set ranges, per-field hash ops and Lua atomicity. All of those are concurrency requirements though, so a single-instance in-memory backend would be straightforward (`Store == nil` instead of `ErrMissingStore`) and would let people try the middleware without running Redis. Better as a follow-up than in this PR.
- `GreenThreshold` defaults to 99.9%. At a 3s interval that is 28,800 slots a day, so 28 missed slots turn a day yellow. One GC pause or Redis hiccup costs a slot.
- The dashboard auto-refreshes and each poll rebuilds the snapshot uncached: `ListServices` plus 30 days × N services, per open tab.
- Endpoint probes run in every replica, so N replicas mean N× probe traffic on the target. Now noted in the README.
- Probes start at `New()`, before `app.Listen()`, so the first probes against your own app always fail. Also noted in the README.

gofurry:
@ReneWerner87  Thanks for testing this against a real Redis instance and for fixing these issues. The `RemoveService` API looks reasonable to me, and I have no objections to the commits pushed to the branch.

gaby (approved):

gaby:
We can add more features later.

PR #1836, opened 2026-04-25
🔥 feat: Add SPNEGO Authentication Middleware

ReneWerner87:
Reopened from #1368 — original was closed due to a broken force-push that orphaned main's history (now restored).

PR #1368, opened 2025-08-06
🔥 feat: Add SPNEGO Authentication Middleware

jarod2011:
## Description
This PR adds enterprise-grade SPNEGO authentication middleware to Fiber, providing Kerberos-based single sign-on (SSO) capabilities. The implementation follows Fiber's middleware patterns with dynamic keytab loading and full RFC compliance.

## Related Issue
Resolves #1367  
Issue: [SPNEGO Middleware Proposal](https://github.com/gofiber/contrib/issues/1367)

## Technical Implementation
- 🛡️ **SPNEGO Protocol**: Full support for `WWW-Authenticate` challenges and `Authorization` validation
- 🔑 **Dynamic Credentials**: Keytab loading via callback functions (files, secrets managers, databases)
- ✅ **RFC Compliance**: Fully implements RFC 4559, RFC 4120, and RFC 7235
- 📊 **Test Coverage**: >90% coverage including edge cases and error handling
- 📚 **Documentation**: Complete usage examples in README
- Support fiber v2 and v3

## Verification
```bash
# Run tests
cd spnego
go test -cover ./...
# Test output
        github.com/gofiber/contrib/spnego               coverage: 0.0% of statements
        github.com/gofiber/contrib/spnego/config                coverage: 0.0% of statements
ok      github.com/gofiber/contrib/spnego/v2    0.389s  coverage: 87.5% of statements
ok      github.com/gofiber/contrib/spnego/v3    0.388s  coverage: 87.5% of statements


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * SPNEGO Kerberos authentication middleware for Fiber v3 with configurable keytab lookup and optional logging
  * Context helpers to store/retrieve authenticated identity

* **Documentation**
  * New README, package docs, and end-to-end example demonstrating setup and usage

* **Utilities**
  * Keytab inspection helpers and a mock keytab builder for tests/examples

* **Tests**
  * Extensive unit and integration tests covering keytab lookup, middleware flows, and utilities
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

PR #2145, opened 2026-08-03
feat: SPNEGO Kerberos Authentication Middleware

### LINKED ISSUE #1367: 📝 [Proposal]: Add Enterprise-Grade SPNEGO Authentication Middleware
### Feature Proposal Description

**Add Enterprise-Grade SPNEGO Authentication Middleware**  
This middleware provides Kerberos-based authentication for Fiber v3:
- Handles full SPNEGO protocol (`WWW-Authenticate` challenges + `Authorization` validation)
- Dynamic keytab loading via callback functions (files, secrets managers, databases)
- Sets authenticated principal in `c.Locals("user")`
- System credential cache fallback
- Automatic SPN/realm resolution
- Customizable unauthorized responses

**Workflow**:
1. Client accesses protected route
2. Middleware returns `401 + WWW-Authenticate: Negotiate`
3. Client provides Kerberos ticket (browser auto-completes)
4. Middleware validates ticket and decrypt request user info
5. Success: Continue route with `c.Locals("user")` set

### Alignment with Express API

This implementation follows Express patterns
```go
authMiddleware, err := NewSpnegoKrb5AuthenticateMiddleware(&Config{
	KeytabLookup: keytabLookup,
})
app.Use("/protected", authMiddleware)
app.Get("/protected/resource", func(c fiber.Ctx) error {
	identity, ok := GetAuthenticatedIdentityFromContext(c)
	if !ok {
		return c.Status(fiber.StatusUnauthorized).SendString("Unauthorized")
	}
	return c.SendString(fmt.Sprintf("Hello, %s!", identity.UserName()))
})
```

### HTTP RFC Standards Compliance

Fully compliant with:
- **RFC 4559** (SPNEGO-based Kerberos HTTP Authentication)
  - Section 4.1: `Negotiate` scheme implementation
  - Section 4.2: GSS-API message encapsulation
- **RFC 4120** (Kerberos V5)
  - AP-REQ/AP-REP validation
  - Service ticket verification
- **RFC 7235** (HTTP/1.1 Authentication)
  - `WWW-Authenticate` challenge generation
  - `Authorization` header processing

**Compliance ensured through**:
- Standardized [gokrb5](https://github.com/jcmturner/gokrb5) implementation
- Test coverage for all authentication flows
- Support for MIT Kerberos and Active Directory

### API Stability

```go

// KeytabLookupFunc is a function type that returns a keytab or an error
// It's used to look up the keytab dynamically when needed
// This design allows for extensibility, enabling keytab retrieval from various sources
// such as databases, remote services, or other custom implementations beyond static files
type KeytabLookupFunc func() (*keytab.Keytab, error)

// Config holds the configuration for the SPNEGO middleware
// It includes the keytab lookup function and a logger
type Config struct {
	// KeytabLookup is a function that retrieves the keytab
	KeytabLookup KeytabLookupFunc
	// Log is the logger used for middleware logging
	Log *log.Logger
}
```

### Feature Examples

```go
This just a example server
1. user should use `kinit` login to kerberos at first and use `klist` check session.
2. use `curl --negotiate ` to trigger spnego authenticate

app := fiber.New()
keytabLookup, err := NewKeytabFileLookupFunc("/keytabFile/one.keytab", "/keytabFile/two.keyta")
if err != nil {
	panic(fmt.Errorf("create keytab lookup function failed: %w", err))
}
authMiddleware, err := NewSpnegoKrb5AuthenticateMiddleware(&Config{
	KeytabLookup: keytabLookup,
})
if err != nil {
	panic(fmt.Errorf("create spnego middleware failed: %w", err))
}
// Apply the middleware to protected routes
app.Use("/protected", authMiddleware)
// Access authenticated identity
app.Get("/protected/resource", func(c fiber.Ctx) error {
    identity, ok := GetAuthenticatedIdentityFromContext(c)
	if !ok {
		return c.Status(fiber.StatusUnauthorized).SendString("Unauthorized")
	}
	return c.SendString(fmt.Sprintf("Hello, %s!", identity.UserName()))
})
log.Info("Server is running on :3000")
go func() {
	<-time.After(time.Second * 1)
	fmt.Println("use curl -kv --negotiate http://sso.example.local:3000/protected/resource")
	fmt.Println("if response is 401, execute `klist` to check use kerberos session")
	<-time.After(time.Second * 2)
	fmt.Println("close server")
	if err = app.Shutdown(); err != nil {
		panic(fmt.Errorf("shutdown server failed: %w", err))
	}
}()
if err := app.Listen("sso.example.local:3000"); err != nil {
	panic(fmt.Errorf("start server failed: %w", err))
}
```

### Checklist:

- [x] I agree to follow Fiber's [Code of Conduct](https://github.com/gofiber/fiber/blob/main/.github/CODE_OF_CONDUCT.md).
- [x] I have searched for existing issues that describe my proposal before opening this one.
- [x] I understand that a proposal that does not meet these guidelines may be closed without explanation.

gaby:
Adds a SPNEGO/Kerberos authentication middleware for Fiber v3 as `v3/spnego`.

Closes #1367. Supersedes #1836 (and #1368 before it).

## Why this PR replaces #1836

#1836 could not be merged. Its branch was created before `main`'s history was force-pushed and restored, so the two shared **no merge base at all** — `git merge-base` returned nothing, and the PR's recorded base commit (`9937cdaa`) no longer exists in the repository. A normal merge refused with `fatal: refusing to merge unrelated histories`, and `--allow-unrelated-histories` produced add/add conflicts on *every file in the repository*, since the merge base was empty.

Rather than resolve thousands of spurious conflicts, the original commits were cherry-picked onto current `main`. **Author attribution is preserved** — the first 15 commits are still authored by @jarod2011 and @gaby, and the `Co-authored-by` trailers for the review bots are intact.

## What the middleware does

- Implements RFC 4559 SPNEGO: issues `WWW-Authenticate: Negotiate` challenges and validates the `Authorization` header.
- Loads keytabs through a `KeytabLookupFunc` callback, so keys can come from files, a secrets manager, or a database rather than only from disk. One or more keytab files can be merged.
- Exposes the authenticated identity to downstream handlers via `GetAuthenticatedIdentityFromContext`.
- Ships `utils` helpers for inspecting keytab principals and encryption types, plus a mock keytab generator used by the tests.

Built on [gokrb5](https://github.com/jcmturner/gokrb5); works against both MIT Kerberos and Active Directory.

## Commits carried over

15 of the 30 commits on the original branch. The other 15 were merge commits from `main` (no content) plus one commit whose entire diff was a single line in `go.work` — see below.

## Changes made while rebasing

Isolated in the commits after the contributors' work, so they can be reviewed separately.

**`go.work` dropped.** This was the only file in #1836 that overlapped with `main`. `main` no longer tracks a workspace file — it is listed in `.gitignore`, and the test workflows set `GOWORK: off` — so the `use ./v3/spnego` entry no longer applies. The one commit that edited nothing else (`58204e34`, "Update go.work") therefore had no content left to apply and was skipped.

**`test-spnego.yml` updated to the current template.** `main`'s test workflows have moved on since #1836 was authored. The workflow now follows the same shape as the other `test-*.yml` files: pinned action SHAs instead of floating tags, `blacksmith-4vcpu-ubuntu-2404`, the shared `gofiber/.github` gotestsum action, a concurrency group, explicit `permissions`, `GOWORK: off`, and `cache-dependency-path`. The Go matrix covers 1.25.x and 1.26.x.

It also carries two steps the other modules do not have. This module has build-tagged sources for file identity (`identity_unix.go`, `identity_windows.go`, `identity_other.go`) and the test job runs on Linux only, so a `go vet` pass over `windows/amd64`, `darwin/arm64` and `js/wasm` keeps the variants CI never executes compiling. And because this middleware decides who is authenticated, a coverage floor fails the build rather than letting a silent drop reach review.

**Listed in `README.md` and `v3/README.md`.** Every other middleware appears in both lists; spnego was missing from them.

**Compile fix in `utils.GetKeytabInfo`.** The module did not build — the stdlib `maps.Values` returns an `iter.Seq`, not a slice, so the two-variable range form is invalid. This predates the rebase; it came in via a review suggestion on the original PR.

**Dependencies updated.** Fiber `v3.0.0-rc.2` → `v3.4.0` and fasthttp `v1.65.0` → `v1.73.0`, bringing the module in line with every other v3 module. Indirect dependencies follow from those. `gokrb5/v8 v8.4.4`, `goidentity/v6 v6.0.1` and `testify v1.11.1` were already current. The rc.2 → v3.4.0 jump required no source changes.

## Review feedback addressed

**Correctness.** `utils.NewMockKeytab` lost the write error whenever the following `Close` succeeded, returning an error that wrapped nil; mock keytab files are now created `0600` rather than `0666`; the keytab lookup error wraps with `%w` so the underlying cause stays reachable via `errors.Is`; testify's `Equal` arguments were in `(actual, expected)` order in 34 assertions.

**Example.** It was `package example` with an `ExampleNew` function in a non-test file, so Go neither treated it as a testable example nor allowed `go run ./example`. It is now `package main` in `main.go`, matching `v3/i18n/example` and `v3/otel/example`. Its mock SPN was also `HTTP/sso1.example.com` while the server listens on `sso.example.local:3000`, so a client's service ticket would never have matched a keytab entry.

## Request path

Both the keytab load and the SPNEGO handler construction originally sat in the per-request path.

`NewKeytabFileLookupFunc` now caches the merged keytab and re-reads the files only when one of them changes size, modification time **or identity**. On Unix identity is the inode, so replacing a keytab by rename is picked up even when a staging tool preserved the timestamp of a same-sized file. On Windows it is the creation time, which NTFS file tunneling can restore and some filesystems do not report at all, so that platform falls back to size and modification time; the same applies anywhere `fs.FileInfo` exposes no identity.

The middleware does **not** use `adaptor.HTTPHandler`. `adaptor.ConvertRequest` parses the client-supplied URI and copies every header on each request, and the adapter runs the wrapped handler on another goroutine. Instead the `net/http` request handed to gokrb5 is built directly, carrying only what gokrb5 v8.4.4 actually reads on the server path — the `Authorization` header and `RemoteAddr`, which it turns into a client-address restriction — and gokrb5's response is captured by a small `http.ResponseWriter` recorder before being replayed onto the Fiber response. The inner handler runs synchronously on the request goroutine, so a panic from a downstream handler keeps its original value and stack.

Fiber hands out header and URI strings as views into fasthttp's request buffer, which is pooled across connections, so the fields a `SessionManager` could retain are copied before the request is built — and only when one is configured, since nothing else outlives the handler.

`KeytabLookup` is still called per request, so lookup errors surface as before and a rotating custom lookup still works. Note that it runs *before* the ticket is examined, so unauthenticated callers drive it too — the README says so, and points implementations at caching plus bounded timeouts.

## Scope gaps from #1367 — now implemented

- **Customizable unauthorized responses.** `Config.Unauthorized` runs when a presented Kerberos ticket is **rejected**. It deliberately does not run for the other two 401s gokrb5 emits — the bare `Negotiate` opening challenge and the continuation token — because clients only begin or continue SPNEGO when those arrive untouched, so letting a handler rewrite them into, say, a 403 would be a permanent silent deny-all.
- **System keytab fallback.** `Config.FallbackToSystemKeytab` and `NewSystemKeytabLookupFunc` load `$KRB5_KTNAME`, else `/etc/krb5.keytab`, accepting MIT's `TYPE:residual` form. The file-backed types (`FILE:`, `WRFILE:`) are stripped; the known non-file types (`KEYRING:`, `MEMORY:`, `DIR:`, `ANY:`, `SRVTAB:`) are rejected at construction rather than treated as a literal path that fails on every request. An empty `Config` still fails as before, so the fallback is opt-in.

  One deliberate deviation: #1367 asked for *credential cache* fallback. A credential cache cannot serve this role — accepting a SPNEGO ticket requires the service's long-term keys, and gokrb5's `SPNEGOKRB5Authenticate` takes a `*keytab.Keytab` with no ccache-based equivalent (`credentials.CCache` is client-side only). The system keytab is the server-side counterpart, and is named accordingly rather than presented as a ccache.

## Configuration added during review

All of it is optional — the zero `Config` behaves exactly as it did before any of this was added.

- **`SessionManager`** wires gokrb5's session support, so an authenticated client is not revalidated on every request. It is the largest saving available on an authenticated hot path, and it changes the trust model: the session becomes a credential in its own right, so what the implementation stores must be unguessable, bound to the client, and expired deliberately. The README documents what the request handed to a manager carries, which parts of its response are replayed, and why the two failure paths are not symmetric — a `New` that cannot persist fails the request, while a failed `Get` degrades silently to full validation.
- **`KeytabPrincipal`** pins which principal's key decrypts a ticket. Without it gokrb5 infers the principal from the ticket itself, so a keytab merged from several services accepts a ticket for any of them — right for one service with one SPN, wrong the moment a keytab is shared. gokrb5's `service.SName` is deliberately *not* exposed: the SPNEGO accept path never reads it, so wiring it up would promise a restriction that does nothing.
- **`MaxClockSkew`**, **`RequireHostAddress`** and **`DisablePACDecoding`** map to the corresponding gokrb5 settings. `DisablePACDecoding` is spelled negatively so the zero value keeps gokrb5's default of decoding, which is what populates the AD group SIDs.
- **`OnSuccess` / `OnError`** are hooks for metrics and audit; neither can change the outcome. `OnError` fires only for the middleware's own failures — a rejected ticket is `Unauthorized`'s business and a challenge is not a failure — and distinguishes the two by sentinel.
- **`Next`** skips the middleware for a request, for exempting health probes or CORS preflights.

**Group-based authorization already worked**, and is now documented rather than added: PAC decoding is on by default, and gokrb5 populates the group SIDs behind `goidentity.Identity.Authorized`, so a handler can check AD group membership without parsing anything itself.

## Operational behaviour worth reviewing

- **Nothing this middleware logs carries key material.** gokrb5's keytab parse errors interpolate the buffer they were handed, and for a single-principal keytab that buffer is the whole key. The loader drops that message and reports the file's name, that it did not parse, and how many bytes were read. An error from a *caller-supplied* `KeytabLookupFunc` cannot be sanitised the same way, so `Config.KeytabLookup` and the README say what not to return and show the safe form. Everything else this package logs that it did not write itself — a session manager's body, a lookup error, a keytab path — is quoted and capped, so a newline in any of it cannot forge a log line under the package's own prefix.
- **A keytab that cannot be read** — deleted, unmounted, permissions revoked — is reported as an error rather than masked by the cache, so revoking a keytab takes effect. A keytab that reads but does not parse is treated as a rotation caught mid-write: the last keytab that parsed cleanly keeps being served for a bounded grace window, after which the error surfaces rather than keeping superseded keys alive indefinitely. Entering, expiring and leaving that state each log once per episode, not once per request.
- **A failed keytab lookup answers a bare `500`** with only `Internal Server Error` in the body, since the detail names keytab paths and OS errors. The detail is logged instead, throttled to one line per 30 seconds so unauthenticated callers cannot drive a log flood. The error still matches the package's sentinels, so an application `ErrorHandler` can identify it — but the status stays `500` even when a custom `KeytabLookupFunc` returns a `*fiber.Error`, so an infrastructure fault can never reach a client as a `401` it would try to answer with credentials.
- **gokrb5's own logging is opt-in.** It has no level of its own and writes a line for every request carrying a token, so `Config.Log` and `UseFiberLogger` default to off.

## Verification

Run against `v3/spnego` with `GOWORK=off`, on **both Go 1.25 and 1.26**:

- `go build ./...` — clean
- `go vet ./...` — clean, and also for `windows/amd64`, `darwin/arm64` and `js/wasm`
- `go test -race -count=3 -shuffle=on ./...` — passing for both `spnego` and `spnego/utils`
- `golangci-lint run --config v3/.golangci.yml` — 0 issues, tests included
- `go mod tidy -diff` — clean
- statement coverage — 100.0%, and the workflow fails the build if it drops below 85%
- `go run ./example` — reaches the listen call (fails only on the DNS name the example documents)

113 tests and 89 subtests cover the three 401 shapes and which of them reaches `Config.Unauthorized`, keytab cache reuse and invalidation, rotation by rename, the degraded/expiry/recovery episode logging and its throttling, system keytab resolution including every residual type, session establishment and replay, identity propagation, and the error-sanitising paths.

The coverage step measures the library packages only. `./example` is a runnable demo with no tests by design, and Go 1.26 folds such packages into the total where 1.25 omits them — including it made the same tree report 100.0% on one matrix entry and 88.0% on the other.

The suite is also checked by mutation: 102 seeded faults across `spnego.go`, `config.go` and `utils` — dropped request fields, flipped status rules, removed lock-discipline steps, widened file permissions, drifted SPNEGO constants, un-sanitised log lines — are each caught by at least one test.


gaby:
We need to add go 1.26.x to the CI tests

gaby:
Pushed `bd20d243` and `32e38c79`.

**Go 1.26.x** added to the test matrix, matching `test-otel.yml` and `test-sentry.yml`. Verified the module builds and passes `go test -race` under both 1.25 and 1.26 locally.

**All five Copilot comments applied:**

- `mock_keytab.go` — keytab files now created `0600` instead of `0666`.
- `mock_keytab.go` — doc example no longer defers `cleanup()` before the `err` check; `cleanup` is nil on the error paths, so the example as written would have panicked.
- `spnego.go` — lookup error wrapped with `%w` instead of `%v`. Confirmed both the sentinel and the underlying error are now reachable via `errors.Is`, and the rendered message is unchanged (so the existing body assertion in `spnego_test.go` still holds).
- testify `Equal` argument order — flagged on two lines, but the same inversion was present in **34 assertions** across `config_test.go`, `identity_test.go`, `utils/keytab_test.go` and `utils/mock_keytab_test.go`. Fixed all of them rather than just the two.

`go build`, `go vet`, `go test -race -count=1 ./...` and `golangci-lint` (0 issues) all clean.


---

gaby:
Pushed `482554d4`, covering the two remaining CodeRabbit threads. Description updated.

**Keytab / handler caching** (`spnego.go#L33-L46`, `config.go#L32-L47`) — both were in the per-request path:

- `NewKeytabFileLookupFunc` now caches the merged keytab and re-reads the files only when one changes size or mtime, so rotating a keytab on disk still takes effect.
- `New` reuses the handler from `adaptor.HTTPHandler(SPNEGOKRB5Authenticate(...))` while the lookup keeps returning the same `*keytab.Keytab`.

To let the handler outlive one request, the adapted `net/http` handler no longer closes over the Fiber context — it reads per-request state from the request context, which works because fasthttp's `RequestCtx` *is* the request's `context.Context`. `KeytabLookup` is still called per request, so lookup errors surface exactly as before and a rotating custom lookup keeps working.

**Scope gaps from #1367** flagged by the Linked Issues check:

- `Config.Unauthorized` customizes the failed-auth response. It runs after the SPNEGO layer writes its challenge, so changing only the status or body leaves the `WWW-Authenticate: Negotiate` flow intact — asserted in the test.
- `Config.FallbackToSystemKeytab` + `NewSystemKeytabLookupFunc` resolve `$KRB5_KTNAME`, else `/etc/krb5.keytab`, accepting MIT's `FILE:` prefix. Opt-in, so an empty `Config` still fails as before.

One deliberate deviation worth a reviewer's eye: #1367 asked for *credential cache* fallback, which isn't implementable here. Accepting a SPNEGO ticket needs the service's long-term keys, and `SPNEGOKRB5Authenticate` takes a `*keytab.Keytab` — gokrb5's `credentials.CCache` is client-side only, with no server-accept equivalent. The system **keytab** is the server-side counterpart, so it's named as such rather than presented as a ccache. Happy to rename or drop it if you'd rather keep this PR to the original scope.

12 new subtests cover cache reuse and invalidation, handler reuse, system keytab resolution, the unauthorized hook, and the opt-in fallback. `go build`, `go vet`, `go test -race`, and golangci-lint (0 issues) all clean on **both Go 1.25 and 1.26**.

Note the pre-merge checks above were computed from an earlier commit — CodeRabbit has been rate-limited since — so the two warnings should clear on its next run.


---

gaby:
@codex review

gaby:
Pushed `2c0a7357`. All four Codex findings were valid; all four are fixed, none declined.

**P1 — empty `MechTypes` panics every protected route.** Reproduced. `SPNEGO.AcceptSecContext` evaluates `NegTokenInit.MechTypes[0]` before `Verify` can reject the mechanisms (`spnego/spnego.go:78` in v8.4.4), and nothing in the unmarshal path requires that sequence to be non-empty — so a base64-valid token with an empty `MechTypes` indexes an empty slice. The request carries no credentials, so this was a one-header DoS against anything behind the middleware.

Not fixed by pre-validating that field. The same unchecked-index pattern appears elsewhere in gokrb5 — `Ticket.GetPACType` indexes `ad2[0]` the same way — so parsing the token here to check one thing would duplicate gokrb5's ASN.1 work on every authenticated request and still cover only the input we already knew about. Instead gokrb5's handler runs under a recover, and a panic out of it is reported as `ErrSPNEGOHandlerFailed`: a sanitised `500`, logged on the existing throttle, passed to `OnError`.

That recover must not swallow the *application's* panics, which this middleware deliberately lets propagate with their own value and stack. So the chain no longer runs inside gokrb5's call — the inner handler records the identity and returns, and `ctx.Next()` runs after `ServeHTTP`. gokrb5 does nothing after `inner.ServeHTTP` on either path that reaches it, so no ordering it depends on changed. Both halves are pinned by tests: the malformed token answers `500`, and a downstream panic still reaches the caller unchanged.

**P2 — nil logger on the PAC path.** Confirmed end to end: `Settings.Logger()` is nil by default, `GetPACType` calls `Printf` on it without checking when a ticket's AD-IfRelevant data fails to unmarshal, and a nil `*log.Logger` does panic. gokrb5 now gets a discarding logger rather than nil when no logging is configured. It costs one formatted string per request thrown away, which is the same work a caller who *did* configure a logger already pays. (The recover above would have caught this too, but a discarded log line is a better outcome than a `500`.)

**P2 — mock keytab permissions.** Correct: `os.OpenFile` applies its mode only when it creates the file, so writing a keytab over a path something else left at `0644` kept it world-readable while looking protected. It is now removed and created exclusively at `0600`, which also leaves no window where the key material is on disk under the wrong mode. A remove that fails for any reason other than "not there" is reported as itself rather than resurfacing as a confusing "file exists".

**P3 — `curl --negotiate` needs `-u :`.** Correct; the example would never have sent the exchange. Fixed.

Verified with a 16-mutant sweep over the new code (no survivors), `-race -count=3 -shuffle=on` on Go 1.25 and 1.26, `GOOS=windows/darwin/freebsd` vet, golangci-lint, and 100% statement coverage.

---

gaby:
@codex review

gaby:
@codex review

ReneWerner87 (approved):

PR #2247, opened 2026-08-23
fix(uptime): let page description use full width

gofurry:
Remove the fixed width limit from the dashboard description so text uses the full card width before wrapping.

## Summary by CodeRabbit

* **Style**
  * Expanded description text to use the available width within uptime dashboard cards.
  * Improved readability for longer descriptions by removing the previous narrow text constraint.


gaby (approved):
LGTM

ReneWerner87 (approved):

PR #2310, opened 2026-09-07
fix(monitor): refine trend chart rendering and interaction

gofurry:
Small follow-up improvements to the monitor dashboard:

- Improve trend chart hover interaction with guides, markers, and clearer tooltips.
- Preserve missing samples as gaps instead of forward-filling or converting them to zero.
- Improve Y-axis scaling and add lightweight time labels.
- Use pointer events for chart interaction.
- Increase the theme toggle hit area without changing its visual size or layout.

### Trend chart interaction

QQ20260907-230835

56e27c54-b0ae-42d8-8b40-f17e6d848635

### Theme toggle hit area

QQ20260907-231753

## Summary by CodeRabbit

- **New Features**
  - Added interactive chart hover guides, highlighted data points, clearer time-axis labels, and improved scaling.
  - Updated chart tooltips with structured marker, name, and value details.
  - Increased the clickable area of the theme control.
  - Added TCP connection counts to Process and System metrics.
  - Preserved compatibility views for legacy snapshot fields.

- **Bug Fixes**
  - Charts now preserve gaps when latency or error data is unavailable.
  - Tooltips report only available data points.


gaby:
@gofurry I also notice the new JSON response from the data breaks old deployments that rely on that data. I think the fields changed ? (This was once the new UI was introduced)

gofurry:
@gaby  You're right, the JSON schema changed when I introduced the redesigned monitor implementation.

The legacy response exposed `pid` / `os`, while the current snapshot uses
`process` / `runtime` / `system` / `http`.

That change was intentional as part of the redesign, but I agree that preserving
compatibility matters more now that monitor is in the 1.x series.

If necessary, I can audit the legacy fields and restore backward compatibility
without removing the new snapshot structure.

gaby:
@gofurry Two things:
- Restore the fields that can be restored.
- Document the JSON response changes in the README for the middleware.

gofurry:
- Restored the legacy `pid` / `os` JSON views while keeping the new
  `process` / `runtime` / `system` / `http` structure as the primary response.
- Restored the legacy fields that can be mapped, including real process/system
  TCP connection counts for `pid.conns` and `os.conns`.
- Documented the legacy-to-primary field mapping and fallback behavior in the
  monitor README.

I also exposed the TCP connection counts in the Process/System metric cards.

The existing monitor tests, race tests, and linter pass.

I also checked the cold snapshot benchmark because connection enumeration adds
some collection cost. On my machine it changed from about 10.35 ms/op to
11.66 ms/op (+1.3 ms, ~12.6%). This remains on the snapshot collection path and
does not affect the application request hot path.

gaby (approved):
👍 LGTM, I started using this new dashboard recently. It is great!

PR #2321, opened 2026-09-09
🐛 fix(websocket): hand over large messages without a second copy, add benchmarks

gaby:
# Description

Two follow-ups to #2320.

**Large messages were copied twice.** The pooled `ReadMessage` from #2320 doubled its buffer while reading and then returned a second, exact-size copy. A message just past 64 MiB therefore held a 128 MiB backing array plus a 64 MiB copy, where the library's `ReadMessage` holds about 80 MiB, so a peer sending oversized messages to a handler without `SetReadLimit` could exhaust memory more cheaply than before. Raised by the Codex security review on #2320 (https://github.com/gofiber/contrib/pull/2320#discussion_r3968534252).

A buffer the pool would not retain is now returned as it is, with no second copy, and past the retained 64 KiB it grows by a quarter instead of doubling, so a large message ends in a buffer at most 25% bigger than itself, the library's own profile. Messages up to 64 KiB keep the pooled buffer and the exact-size copy, so the fast path is unchanged.

**The middleware had no benchmarks.** `v3/websocket/benchmark_test.go` covers the basic paths so improvements can be measured going forward: the handshake without a socket (accepted, rejected, not an upgrade), `IsWebSocketUpgrade` and the `Conn` accessors; a full upgrade over loopback; echo at 5 B, 1, 16 and 64 KiB and pipelined echo with sixteen frames per write; push from the handler; the library's `ReadMessage` against the pooled one; and one exchange through the write coalescer alone. The file is self-contained and touches no existing test.

Fixes: no linked issue.

## Changes introduced

- [x] Benchmarks: a 64 MiB + 1 byte message through `ReadMessage`, measured with `runtime.MemStats`:

  | | held when `ReadMessage` returns |
  |---|---|
  | before | 128 MiB backing array + 64 MiB copy |
  | after | one 83 MB buffer (1.23× the message), no copy |

  The new benchmarks on a 4-core machine, for orientation: `Upgrade/accepted` 7.0 µs and 68 allocs, `Upgrade/notUpgrade` 365 ns and 0 allocs, `Echo/5B` 10.6 µs per round trip, `EchoPipelined` 1.4 µs per frame, `ReadMessage/library/16KB` 27.3 µs and 14 allocs against `ReadMessage/pooled/16KB` 11.8 µs and 2 allocs, `CoalescingConn/1frames` 105 ns and `CoalescingConn/16frames` 543 ns with 0 allocs. The loopback ones include the client and the kernel, so they are for comparing commits on one machine.
- [x] Documentation Update: `v3/websocket/README.md`, the *Reads and writes* paragraph now says `ReadMessage` returns an exact-size copy up to 64 KiB and the message's own buffer above that, and points at `SetReadLimit` as the bound on message size.
- [x] Changelog/What's New:
  - websocket: `(*Conn).ReadMessage` no longer copies messages larger than 64 KiB; a large message's memory is bounded to 1.25× its size, as with the library's `ReadMessage`.
  - websocket: benchmarks for the upgrade, echo, pipelined echo, push, `ReadMessage` and the write coalescer (`go test -run '^$' -bench . -benchmem` in `v3/websocket`).
- [x] Migration Guide: no code changes needed. A `ReadMessage` result larger than 64 KiB now has capacity above its length, up to 25%, where it was exact before; code that appends to the result sees the same bytes either way.
- [ ] API Alignment with Express: not applicable.
- [x] API Longevity: no API change. `ReadMessage` keeps the library's signature and the caller still owns the returned slice.
- [x] Examples: `go test -run '^$' -bench . -benchmem` in `v3/websocket` runs the whole set; `-bench ReadMessage` compares the two readers.

## Type of change

- [x] Code consistency (non-breaking change which improves code reliability and robustness)
- [x] Documentation update (changes to documentation)

## Checklist

- [ ] Followed the inspiration of the Express.js framework for new functionalities, making them similar in usage. (Not applicable: no new API.)
- [x] Conducted a self-review of the code and provided comments for complex or critical parts.
- [x] Updated the documentation in the module README (this repository has no `/docs/` directory).
- [x] Added or updated unit tests to validate the effectiveness of the changes or new features: the reader test asserts the bounded growth and the hand-over, and the owned-copy test checks that a large result survives the reads that follow it.
- [x] Ensured that new and existing unit tests pass locally with the changes (`go test -race ./...` in `v3/websocket`, both packages).
- [x] Verified that any new dependencies are essential and have been agreed upon by the maintainers/community. (No new dependencies; `go.mod` is unchanged.)
- [x] Aimed for optimal performance with minimal allocations in the new code.
- [x] Provided benchmarks for the new code to analyze and improve upon.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MqWdVBHdP7GVYYB7o5MHVS

---
_Generated by [Claude Code](https://claude.ai/code/session_01MqWdVBHdP7GVYYB7o5MHVS)_

## Summary by CodeRabbit

- **Performance**
  - Improved WebSocket message handling to reduce unnecessary memory copying, especially for messages larger than the retained buffer size.
  - Preserved message ownership across successive reads so previously returned message data remains stable.

- **Documentation**
  - Clarified WebSocket message buffering behavior, handling of large messages, and how to enforce maximum message sizes with read limits.

### DIFF
--- a/v3/websocket/README.md
+++ b/v3/websocket/README.md
@@ -128,9 +128,10 @@ by earlier middleware are left in place.
 
 ## Reads and writes
 
-`c.ReadMessage()` reads into a pooled buffer and returns an exact-size copy, where the
-library's `ReadMessage` grows a fresh buffer per message; for zero-copy reads use
-`NextReader` with your own buffer.
+`c.ReadMessage()` reads messages up to 64 KiB into a pooled buffer and returns an exact-size
+copy; a larger message comes back in its own buffer, as from the library's `ReadMessage`.
+Bound message size with `SetReadLimit`; for zero-copy reads use `NextReader` with your own
+buffer.
 
 Replies to a burst of pipelined frames leave in one write instead of one per frame: writes
 made while the handler still has unread frames wait until it asks for the next one, at most
--- a/v3/websocket/benchmark_test.go
+++ b/v3/websocket/benchmark_test.go
@@ -1,0 +1,377 @@
+package websocket
+
+import (
+	"bufio"
+	"fmt"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/fasthttp/websocket"
+	"github.com/gofiber/fiber/v3"
+	"github.com/stretchr/testify/require"
+	"github.com/valyala/fasthttp"
+)
+
+// The middleware's own cost, from the router to the socket. Run with
+//
+//	go test -run '^$' -bench . -benchmem
+//
+// BenchmarkUpgrade and BenchmarkCoalescingConn measure the middleware alone.
+// The loopback benchmarks include the client and the kernel, so compare them
+// across commits on one machine rather than against other servers.
+
+var payloadSizes = []struct {
+	name string
+	size int
+}{{"5B", 5}, {"1KB", 1 << 10}, {"16KB", 16 << 10}, {"64KB", 64 << 10}}
+
+func patterned(size int) []byte {
+	p := make([]byte, size)
+	for i := range p {
+		p[i] = byte(i % 251)
+	}
+	return p
+}
+
+// BenchmarkUpgrade is the handshake without a socket: the router, the
+// middleware and the library's Upgrade, up to the 101 fasthttp would send.
+func BenchmarkUpgrade(b *testing.B) {
+	app := fiber.New()
+	app.Get("/ws", New(func(*Conn) {}))
+	h := app.Handler()
+	run := func(name string, fctx *fasthttp.RequestCtx, status int) {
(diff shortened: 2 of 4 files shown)


gaby:
`lint / changes` failed before any lint ran: `dorny/paths-filter` got a `Connect Timeout Error` from the GitHub API while listing this PR's files, so the lint matrix was never built. Nothing in the diff is involved; the same reusable workflow passed on `main` twenty minutes earlier. Re-running the failed job once. Tests, CodeQL and the labeler are green on 623a52b.

---

ReneWerner87 (approved):

